PR #1, opened 2022-07-12
Fix the instructions for cloning the repo

alt-graph:
[why]
The old instructions still referred to the defunct Stash repo.

[how]
Use the new GitHub HTTPS URL. Also add some minor improvements to the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,8 +8,10 @@ by the library are:
   * Concurrency
   * Time
   * Statistics and numerics
-  * Containers
   * Debugging
+  * Containers
+  * Type traits
+  * Core guideline support library (GSL) excerpts
   * Backports from new standard library extensions
 
 To get started, have a look at the [documentation website](https://winweb.desy.de/mcs/docs/gul/).
@@ -31,8 +33,8 @@ Here you find more detailed information on:
 
 Clone this repository:
 
-        git clone ssh://git@stash.desy.de:7999/gul/libgul.git
-        cd libgul
+        git clone https://github.com/gul-cpp/gul14.git
+        cd gul14
 
 Use meson to configure the build you want to have. A list of the
 [build options](#Build-configuration-switches) is given below. Build directory name can be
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -31,15 +33,15 @@ Here you find more detailed information on:
...
+        git clone https://github.com/gul-cpp/gul14.git
+        cd gul14
 
 Use meson to configure the build you want to have. A list of the
 [build options](#Build-configuration-switches) is given below. Build directory name can be
 what you want. If you use directories starting with 'build' they will be ignored by git.
 Examples:
 
-        meson build
+        meson builddir
```
soerengrunewald, line 44: Maybe we should use `<builddir>` instead, since one can also use e.g. `/tmp/build` which is on modern linux distros in memory.

README.md
```diff
@@ -51,12 +53,12 @@ Typical build configurations can be easily summoned as detailed in
 
 Afterwards enter you build directory and fire the build process:
 
-        cd build_rel
+        cd builddir
```
soerengrunewald, line 56: Same here

README.md
alt-graph: But since this is an example, people probably expect to be able to copy&paste it directly, no? It is not a command synopsis or anything like that.

Finii: But why `builddir`? In the next lines it is always varying...
* `build_without_tests`
* `build_rel`
* `build_rel_unity`

README.md
Finii: Well, this change is (clearly?) wrong, because the release build process shall be initiated, and some lines above the matching example uses the directory `build_rel`, and not `builddir`, so this becomes disconnected.

I, personally, hate `builddir`, and maybe how VScode handles this. I always have multiple build dirs set up; at least one for GCC, one for clang, one for release.

Finii (changes requested):

README.md
alt-graph: If we want to explain how to make a release build, I guess we should say so. I'll adapt the text accordingly. And since Fini hates `builddir`, I'll revert that to build.

Finii:
> But makeDdeb is a DOOCS script or not?

Yes. It is basically the MCS version of `debuild` (which is the 'official Debian package build tool) - both rely on `dpkg-buildpackage`. I usually use `debuild` to test the GUL build process.

Both rely on our `debian/rules` to know what to do.

README.md
alt-graph: Okay, back to `build` it is.

Finii (approved):

alt-graph:
> But makeDdeb is a DOOCS script or not?

That is a good point. I tried to make it less confusing for non-DESY readers.

PR #4, opened 2022-07-18
Documentation: Add info on vcpkg installation

alt-graph:
vcpkg has recently accepted a [portfile for GUL14](https://github.com/microsoft/vcpkg/commit/fd69fb296475fb8c5ef5e3daec57258e5c7ba06d), so the library can now be installed via "vcpkg install gul14". It is probably good to advertise this in a new "installation" section in both the README and the Doxygen documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,12 @@ Here you find more detailed information on:
 * [Versioning](#Versioning)
   * [Version number examples](#Version-number-examples)
 
+## Installing with vcpkg
+
+GUL14 is available from the [vcpkg](https://vcpkg.io/) package manager. Once you have
+vcpkg installed, just run:
+
+        vcpkg install gul14
 
 ## Building <a name="Building"></a>
 
--- a/data/doxygen.h
+++ b/data/doxygen.h
@@ -79,6 +79,13 @@ namespace gul14 {
  * GUL14 requires at least C++14. It works fine with newer versions of the standard, but
  * uses its own backport types (e.g. gul14::string_view) in function interfaces.
  *
+ * \section installation Installation
+ *
+ * If you are using the [vcpkg](https://vcpkg.io/) package manager, you can install the
+ * library simply by running \verbatim vcpkg install gul14\endverbatim. Otherwise, you may
+ * have to build and install it manually. Have a look at the
+ * [readme file on GitHub](https://github.com/gul-cpp/gul14/blob/main/README.md).
+ *
  * \section source_code Obtaining the Source Code
  *
  * You can browse or clone the source code at https://github.com/gul-cpp/gul14.git.


Finii:
The vcpkg-file looks a bit strange for me, as it uses `https://github.desy.de/gul-cpp/gul14.git` ?

Also, what I see now, the Doxygen stuff is hosted at Desy, would it not make more sense to but it into `gh-pages` in the github repo?

As for this PR, I guess it is ok, although I have zero experience with (and zero inclination to use) vcpkg.

But as GUL is built with Meson, we might (also) add it to the wrap db.

Finii (approved):
Maybe you could show what happens after `vcpkg install`, just that I/we can understand what it does. Not that this has anything to do directly with this PR, and is independent.

Finii:
> Also, what I see now, the Doxygen stuff is hosted at Desy, would it not make more sense to but it into `gh-pages` in the github repo?

Did it: https://gul-cpp.github.io/gul14/

Will change README without PR ;)

soerengrunewald (approved):

Finii:
> https://gul-cpp.github.io/gul14/

Added a small manual, for when we release...

https://github.com/gul-cpp/gul14/tree/gh-pages

alt-graph:
> But as GUL is built with Meson, we might (also) add it to the wrap db.

Adding GUL14 to the Wrap DB sounds like a great idea! 👍

alt-graph:
> Maybe you could show what happens after `vcpkg install`, just that I/we can understand what it does. Not that this has anything to do directly with this PR, and is independent.

Well, not a lot. `vcpkg` compiles the library and all prerequisites and then installs all of its libraries, headers and pkg-config files into a certain directory, in my case the pkg-config stuff goes into `/home/lfroehli/vcpkg/installed/x64-linux/lib/pkgconfig`. You simply add this to your PKG_CONFIG_PATH and you are done. For Visual Studio and cmake there is even deeper integration, but whatever...

alt-graph:
> > Also, what I see now, the Doxygen stuff is hosted at Desy, would it not make more sense to but it into `gh-pages` in the github repo?
> 
> Did it: https://gul-cpp.github.io/gul14/

Well, we are in the process of getting a subdomain at DESY: https://gul14.desy.de
Maybe that is a nicer URL and/or looks more professional?

Finii:
> Well, we are in the process of getting a subdomain at DESY: https://gul14.desy.de Maybe that is a nicer URL and/or looks more professional?

I don't know. Still one can host it at Github nevertheless (Github allows custom domains).
Must confess I like it more if everything is in one place, in contrast to having unknow-to-other stuff has to run to update the web pages. If it's on Github the pages can automatically be updated by the Release-CI (that we do not have, yet).

'Typically' on Github I do this: If a version tag is pushed the CI triggers a new release, which is build and published, together with updated gh-pages. So from a developer standpoint that is zero work, just set the git tag and push.

Finii:
> OK... I could not resist and registered https://www.gul14.info

Updated the link on the github front page (top right)

image

PR #9, opened 2022-07-29
Add date header

alt-graph:
This PR adds Howard Hinnant's [date.h](https://github.com/HowardHinnant/date) header as proposed in #7. I have made the "usual" modifications for GUL14: The header is enclosed in namespace `gul14::date::` and it uses `gul14::string_view` instead of `std::string_view`. I also added the necessary pieces of documentation and copyright notices and tentatively increased our version number to 2.7.

### DIFF
--- a/data/doxygen.h
+++ b/data/doxygen.h
@@ -70,7 +70,8 @@ namespace gul14 {
  *
  * To use the library:
  *
- *  - Include the main header file \ref gul14/gul.h.
+ *  - Include the main header file \ref gul14/gul.h (or read more under
+ *    \ref how_to_include_gul_headers).
  *  - Link your code against the library (-lgul14).
  *  - If you are on Windows, see \ref using_gul_on_windows.
  *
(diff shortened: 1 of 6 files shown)


alt-graph:
One point we should discuss before merging: The header is fairly big and full of template definitions. I assume it will be somewhat costly in terms of compile time. Therefore, **should we include <gul14/date.h> by default with <gul14/gul.h>?** This PR currently does so, but personally I tend towards no.

Finii:
> but personally I tend towards no.

I would also _not_ include it.

Finii (commented):
Seems that the date header is not installed?

Finii:
I usually first add the unmodified file as one commit and then add our specific changes as separate commit; then one can see what has been customized. Also helpful if the upstream source is updated and one wants to follow that, often the same diff can be used easily.

Did not see anything to complain about apart from the install.

alt-graph:
> Seems that the date header is not installed?

Ouch, thanks! Done. :)

alt-graph:
Fixed the missing header installation.
I also added a bit of Doxygen documentation on how to include GUL14's headers.

Finii (approved):

PR #10, opened 2022-08-05
Add repeat() function to repeat a string N times

alt-graph:
Add repeat() to repeat a string _N_ times
    
The C++ standard library has no convenient function to repeat a string N times, but it is occasionally useful:
```C++
auto str = repeat("du", 3); // returns "dududu"
```

### DIFF
--- a/src/string_util.cc
+++ b/src/string_util.cc
@@ -4,7 +4,7 @@
  * \authors \ref contributors
  * \date    Created on 31 August 2018
  *
- * \copyright Copyright 2018-2021 Deutsches Elektronen-Synchrotron (DESY), Hamburg
+ * \copyright Copyright 2018-2022 Deutsches Elektronen-Synchrotron (DESY), Hamburg
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU Lesser General Public License as published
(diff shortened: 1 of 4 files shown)


src/string_util.cc
```diff
@@ -29,6 +29,17 @@ const string_view default_whitespace_characters{ " \t\r\n\a\b\f\v" };
 const std::array<char, 16> hex_digits{
     { '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'a', 'b', 'c', 'd', 'e', 'f'} };
 
+std::string repeat(gul14::string_view str, std::size_t n)
+{
+    std::string result;
+
```
Finii, line 35: Maybe a small optimization, usually Lars' job to suggest these ;-)

```suggestion
    if (n == 0 or str.empty())
        return "";
```

Finii (approved):
Tried to find out what other languages call this...

Java(Script): `str.repeat()`
Puthon: `operator*`
Perl: `operator_x` :grimacing:

Reminds me of my Doocs [talk](https://confluence.desy.de/download/attachments/185871363/serial_com_2.pdf?version=1&modificationDate=1619593504215&api=v2) last year... (box on rhs):

image

src/string_util.cc
alt-graph: Hm... that raises the number of lines of x86-64 assembly from 54 to 82 on Godbolt (gcc 12.1), and if I understand correctly it disables NRVO. Does it improve performance for these corner cases? Maybe, but we'd need benchmarks to prove that. I would rather leave it out.

Finii: I'm fine with that.

I already approved.

PR #11, opened 2022-08-08
Add gcd() and lcm()

alt-graph:
These functions can calculate the [greatest common divisor](https://en.wikipedia.org/wiki/Greatest_common_divisor) and [least common multiple](https://en.wikipedia.org/wiki/Least_common_multiple) of two integers.

They are very similar to the C++17 functions [std::gcd()](https://en.cppreference.com/w/cpp/numeric/gcd) and [std::lcm()](https://en.cppreference.com/w/cpp/numeric/lcm), but do not work with integers of different signedness – a static_assert prevents that. This makes them safer to use because the standard versions cause UB when used with one unsigned integer and one negative integer.

### DIFF
--- a/include/gul14/gcd_lcm.h
+++ b/include/gul14/gcd_lcm.h
@@ -23,6 +23,7 @@
 #ifndef GUL14_GCD_LCM_H_
 #define GUL14_GCD_LCM_H_
 
+#include <cstdint>
 #include <type_traits>
 #include "gul14/internal.h"
 #include "gul14/num_util.h"
@@ -129,7 +130,7 @@ lcm(IntTypeA a, IntTypeB b)
     if (a == 0 && b == 0)
         return CommonType{ 0 };
 
-    return static_cast<CommonType>(gul14::abs(a * b) / gcd(a, b));
+    return static_cast<CommonType>(gul14::abs((a / gcd(a, b)) * b));
 }
 
 } // namespace gul14
--- a/include/gul14/gcd_lcm.h
+++ b/include/gul14/gcd_lcm.h
@@ -1,6 +1,6 @@
 /**
  * \file    gcd_lcm.h
- * \brief   Declarations of greatest_common_divisor() and least_common_multiple().
+ * \brief   Declarations of gcd() and lcm().
  * \authors \ref contributors
  * \date    Created on 5 August 2022
  *
@@ -23,7 +23,6 @@
 #ifndef GUL14_GCD_LCM_H_
 #define GUL14_GCD_LCM_H_
 
-#include <cstdint>
 #include <type_traits>
 #include "gul14/internal.h"
 #include "gul14/num_util.h"
@@ -31,8 +30,7 @@
 namespace gul14 {
 
 /**
- * Calculate the greatest common divisor (gcd) of two integers using the Euclidean
- * algorithm.
+ * Calculate the greatest common divisor of two integers using the Euclidean algorithm.
  *
  * If both numbers are zero, the function returns zero. Otherwise, the result is a
  * positive integer.
--- a/include/gul14/gcd_lcm.h
+++ b/include/gul14/gcd_lcm.h
@@ -1,0 +1,137 @@
+/**
+ * \file    gcd_lcm.h
+ * \brief   Declarations of greatest_common_divisor() and least_common_multiple().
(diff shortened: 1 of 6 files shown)


include/gul14/gcd_lcm.h
```diff
@@ -0,0 +1,137 @@
...
+
+    static_assert(std::is_signed<IntTypeA>::value == std::is_signed<IntTypeB>::value,
+        "lcm() arguments must have the same signedness");
+
+    using CommonType = std::common_type_t<IntTypeA, IntTypeB>;
+
+    if (a == 0 && b == 0)
+        return CommonType{ 0 };
+
+    return static_cast<CommonType>(gul14::abs(a * b) / gcd(a, b));
```
Finii, line 132: This can overflow for smallish `IntType` and / or moderate values of `a` and `b`. But the docu does not say so.
It might be that it works for `unsigned`, because it will just owerflow the other side again, but for signed integers overlow is UB, iirc.

(Because `a * b`.)

tests/test_gcd_lcm.cc
```diff
@@ -0,0 +1,133 @@
...
+    REQUIRE(lcm(15, -10) == 30);
+
+    REQUIRE(lcm(7, 5) == 35);
+    REQUIRE(lcm(-7, 5) == 35);
+    REQUIRE(lcm(7, -5) == 35);
+    REQUIRE(lcm(-7, -5) == 35);
+
+    REQUIRE(lcm(std::int16_t{ -256 }, std::int8_t{ -128 }) == std::int16_t{ 256 }); // Fails with GCC 9.3.0 std::lcm()
+    REQUIRE(lcm(-288LL, std::int8_t{ -128 }) == 1152LL);
+
```
Finii, line 107:
```suggestion
   REQUIRE(lcm(6, 10) == 30);
    REQUIRE(lcm(60, 100) == 300);
    REQUIRE(lcm(600, 1'000) == 3'000);
    REQUIRE(lcm(6'000, 10'000) == 30'000);
    REQUIRE(lcm(60'000, 100'000) == 300'000);
```

Last case fails on doocsdev20 for example.

Finii (changes requested):

include/gul14/gcd_lcm.h
Finii: Also fails for `unsigned`

```
../tests/test_gcd_lcm.cc:113: FAILED:
  REQUIRE( lcm(60'000u, 100'000u) == 300'000u )
with expansion:
  85251 (0x14d03) == 300000 (0x493e0)
```

tests/test_gcd_lcm.cc
Finii: I have no easy solution for that; maybe it would be good to switch to prime factorization method instead of GCF method.
:woman_shrugging:

include/gul14/gcd_lcm.h
alt-graph: Very good catch, thanks! I added a commit that uses `intmax_t`/`uintmax_t` for the product and adds documentation on UB should the product not be representable by that type.

tests/test_gcd_lcm.cc
alt-graph: I have added these tests and a few more. They now run cleanly.
But I agree that this can't be the solution. We should be able to calculate
```
lcm(6'000'000'000LL, 10'000'000'000LL)
```
correctly as well, but the product overflows intmax_t.

alt-graph: Oh, got it! It is sufficient to change the order of the operations, of course. From the commit message:

    Avoid integer overflow by changing calc order
    
    [why]
    The lcm was calculated as (a * b) / gcd(a, b).
    The product a * b could easily overflow in the
    CommonType of a and b.
    
    [how]
    Calculate lcm as (a / gcd(a, b)) * b. The division
    is guaranteed to give the correct result because
    gcd is a divisor of a by definition.

That should solve it. I also added test cases where the product would overflow int64_t.

include/gul14/gcd_lcm.h
alt-graph: Found an even better solution, see next thread.

Finii (approved):
Great solution :-) I was blind ;-)

PR #14, opened 2022-09-02
Make sleep() time tests more robust

alt-graph:
This MR increases the tolerance for time measurements after `sleep()` in unit tests. It also improves the documentation of several `sleep()` overloads to make it clear that the functions introduce a delay that is *at least* as long as specified.
    
Many `sleep()` related tests fail on GitHub MacOS runners because the systems are under heavy contention. As `sleep()` is an interruption point, context switches sometimes lead to delays on the order of several tens of milliseconds. We adapt to that by increasing the tolerance after `sleep()` from 18 to 100 ms.

### DIFF
--- a/include/gul14/time_util.h
+++ b/include/gul14/time_util.h
@@ -4,7 +4,7 @@
  * \date    Created on September 7, 2018
  * \brief   Declaration of time related functions for the General Utility Library.
  *
- * \copyright Copyright 2018-2020 Deutsches Elektronen-Synchrotron (DESY), Hamburg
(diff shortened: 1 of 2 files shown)


Finii:
I would
* Just include tests that check what we guarantee (is later than)
* Make other tests (not later than) optional

In this way we would test always only what we really guarantee. But can enable the additional tests on platforms where we are confident that they are fast enough.

alt-graph:
I am not a big fan of optional tests. Strictly speaking, we should only test what we guarantee. There may always be an occasion in which a stupid context switch takes a second. I'll remove the "not-later-than" tests.

Finii (approved):

PR #15, opened 2022-09-27
Fix Trigger wait tests

alt-graph:
This PR removes several "not-later-than" tests for Trigger::wait_for() and Trigger::wait_until().
    
Many Trigger-related tests fail on GitHub MacOS runners because the systems are under heavy contention. This sometimes leads to delays on the order of several tens of milliseconds, which is outside of the tolerance of the old delay tests.
    
Strictly speaking, there is no guarantee that Trigger's wait* functions wake up after any particular time span at all, so we should not test for that.

Drive-by-corrections include documentation and whitespace. :)

### DIFF
--- a/include/gul14/Trigger.h
+++ b/include/gul14/Trigger.h
@@ -4,7 +4,7 @@
  * \date    Created on September 21, 2018
  * \brief   Declaration of the Trigger class for the General Utility Library.
  *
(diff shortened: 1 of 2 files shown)


alt-graph:
This fixes the last failing tests in the pipeline. Should we call it version 2.7 after this is merged?

Finii (approved):

Issue #17, opened 2022-10-11
within_orders() with unexpected result


Finii:
```
  REQUIRE( gul14::within_orders(0.0, 0.0, 5) == true )
with expansion:
  false == true
```

```
  REQUIRE( gul14::within_orders<double>(0.0, 0.0, 5) == true )
with expansion:
  false == true
```

```c++
    double a = 0.0;
    REQUIRE(gul14::within_orders<double>(a, a, 5) == true);
```
```
  REQUIRE( gul14::within_orders<double>(a, a, 5) == true )
with expansion:
  false == true
```

WAT?

Finii:
This is funny @soerengrunewald ;-)

Finii:
Ok, it is **documented**

```
 * Remember that any nonzero number has infinite different significant digits compared
 * with 0.00000000. So if either a or b is 0.0 the result must be false.
```

But still, this feels wrong.

PR #18, opened 2022-10-11
num_util: within_orders() detects 0.0 == 0.0

Finii:
**[why]**
`gul14::within_orders(0.0, 0.0, 5)` is `false`. Instinctively one might expect it to be true, as both inputs are strictly equal.

**[how]**
Check for equality first, then additionally add some lenience.

Fixes: #17

### DIFF
--- a/data/doxygen.h
+++ b/data/doxygen.h
@@ -143,6 +143,11 @@ namespace gul14 {
  *
  * \section changelog_2_x 2.x Versions
  *
+ * \subsection v2_7_1 Version 2.7.1
+ *
+ * - *Released with DOOCS 22.10.0*
+ * - within_orders(0.0, 0.0, 10) results now in true
+ *
  * \subsection v2_7 Version 2.7
  *
  * - Add repeat()
--- a/include/gul14/gcd_lcm.h
+++ b/include/gul14/gcd_lcm.h
@@ -57,7 +57,7 @@ namespace gul14 {
  * auto static_assertion_failure = gul14::gcd(10u, -5);
  * \endcode
  *
- * \since GUL 2.7
+ * \since GUL version 2.7
  */
 template <typename IntTypeA, typename IntTypeB>
 constexpr inline auto
@@ -111,7 +111,7 @@ gcd(IntTypeA a, IntTypeB b)
  * auto static_assertion_failure = gul14::lcm(10u, -5);
  * \endcode
  *
- * \since GUL 2.7
+ * \since GUL version 2.7
  */
 template <typename IntTypeA, typename IntTypeB>
 constexpr inline auto
--- a/include/gul14/num_util.h
+++ b/include/gul14/num_util.h
@@ -71,16 +71,17 @@ constexpr auto abs(ValueT n) noexcept -> std::enable_if_t<not std::is_unsigned<V
  * orders equals digits does not hold so strict anymore.
  *
  * Remember that any nonzero number has infinite different significant digits compared
- * with 0.00000000. So if either a or b is 0.0 the result must be false.
+ * with 0.00000000. So if one operand is 0.0 while the other is not 0.0 the result must be false.
  *
  * \param a       The first number to compare
- * \param b       The second number to compare (same type as a)
+ * \param b       The second number to compare (same type as \c a)
  * \param orders  The number of digits to take for comparison (any numeric type)
  *
- * \returns true if the difference between a and b is orders of magnitude lower than the
- *          value of a or b.
+ * \returns true if \c a and \c b are equal or the difference between \c a and \c b is
+ *          \c orders orders of magnitude lower than the value of \c a or \c b
  *
  * \since GUL version 1.4 parameter type \b NumT can be an integral type (had to be floating point before)
+ * \since GUL version 2.7.1 return true if a == b == 0.0
  */
 template<typename NumT, typename OrderT,
     typename = std::enable_if_t<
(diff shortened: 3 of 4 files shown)


Finii:
Alternative implementation, which uses ULPs.

In fact we could have used `return difference <= limit;` BUT that breaks with special floating point values.
So it's either the PR version, or this:

```c++
bool within_orders(const NumT a, const NumT b, const OrderT orders) noexcept(false) {
    // std::pow() is not noexcept, which might or might not be true
    auto difference = gul14::abs(a - b);
    //if (difference == NumT{ 0 })
      //  return true;
    auto maximum = std::max(gul14::abs(a), gul14::abs(b));
    auto limit = static_cast<NumT>(maximum / std::pow(static_cast<std::decay_t<NumT>>(10.0), orders));
    auto ulp = std::numeric_limits<NumT>::min();
    return difference < std::max(limit, ulp);
}
```

include/gul14/num_util.h
```diff
@@ -71,7 +71,7 @@ constexpr auto abs(ValueT n) noexcept -> std::enable_if_t<not std::is_unsigned<V
  * orders equals digits does not hold so strict anymore.
  *
  * Remember that any nonzero number has infinite different significant digits compared
- * with 0.00000000. So if either a or b is 0.0 the result must be false.
+ * with 0.00000000. So if either a or b (not both) is 0.0 the result must be false.
```
alt-graph, line 74: Maybe we should also spell out explicitly that the function returns true if a == b == 0.

tests/test_num_util.cc
```diff
@@ -146,6 +146,11 @@ TEST_CASE("test within_orders()", "[num_util]")
     REQUIRE(gul14::within_orders(1.0, 0.0, 1) == false);
     REQUIRE(gul14::within_orders(1.0, 0.0, 0) == false);
 
+    // Compare zero with zero
+    double a = 0.0;
+    REQUIRE(gul14::within_orders<double>(a, a, 5) == true);
+    REQUIRE(gul14::within_orders(static_cast<double>(0.0f), 0.0, 5) == true);
```
alt-graph, line 152: The static_cast strikes me as weird: Why not directly test `within_orders(0.0, 0.0, 5) == true`?

alt-graph (changes requested):
Good solution, I would just recommend to update the documentation a little more.

alt-graph:
Before merging, maybe we should also bump the version number?

tests/test_num_util.cc
Finii: I wanted to have a cast from `float` to `double`, as that changes the internal representation and thus the value (sometimes).

Of course a zero should still be a zero ofterwards.

Maybe reflects better my intentions:

```suggestion
    float b = 0.0f;
    REQUIRE(gul14::within_orders(static_cast<double>(b), 0.0, 5) == true);
```

But then, maybe just rely on the fact that a `float` zero cast to `double` is the same as a direct `double` zero is.

This is not the case for some other numbers like 0.1, as they are represented with different rounding.
I.e. `0.1f != 0.1`.

include/gul14/num_util.h
Finii: Done

tests/test_num_util.cc
Finii: Changed code a bit.

Finii:
Increased version number and added git tag (this will work if we fast-forward merge) (hopefully ;-)

Finii:
Obviously 'dismiss' is not the correct button :grimacing:

Finii:
> Before merging, maybe we should also bump the version number?

Hmm, there is no Doocs release between 2.7 and now, so maybe keep 2.7?

image

Finii:
Ah, more appropriate would be `v2.7.1` which keeps the API version at `2.7`... changing

Finii:
image

data/doxygen.h
```diff
@@ -145,9 +145,11 @@ namespace gul14 {
  *
  * \subsection v2_7 Version 2.7
  *
+ * - *Released with DOOCS 22.10.0*
  * - Add repeat()
  * - Add gcd() and lcm()
  * - Add Howard Hinnant's date.h header
+ * - within_orders(0.0, 0.0, 10) results now in true
```
alt-graph, line 152: I agree with going from a 2-part version number to a 3-part one (2.7 -> 2.7.1), but then IMO we should also list the changes by minor version number.

include/gul14/gcd_lcm.h
```diff
@@ -57,7 +57,7 @@ namespace gul14 {
  * auto static_assertion_failure = gul14::gcd(10u, -5);
  * \endcode
  *
- * \since GUL 2.7
+ * \since GUL version 2.7
```
alt-graph, line 60: Good catch!

include/gul14/num_util.h
```diff
@@ -71,16 +71,17 @@ constexpr auto abs(ValueT n) noexcept -> std::enable_if_t<not std::is_unsigned<V
...
- * with 0.00000000. So if either a or b is 0.0 the result must be false.
+ * with 0.00000000. So if one operand is 0.0 while the other is not 0.0 the result must be false.
  *
  * \param a       The first number to compare
  * \param b       The second number to compare (same type as a)
  * \param orders  The number of digits to take for comparison (any numeric type)
  *
- * \returns true if the difference between a and b is orders of magnitude lower than the
- *          value of a or b.
+ * \returns true if a and b are equal or the difference between a and b is orders of
```
alt-graph, line 80: Maybe we could also make the end of the sentence more explicit – I always fail to read it in the intended way: ... is \c orders orders of magnitude lower ... (or add quotation marks or whatever)

alt-graph (approved):
👍

include/gul14/num_util.h
Finii: Hmm, the `\c` does switch to a monospaced font (I guess), but on my machine the difference is almost invisible.
Would have liked a colored background or something. Maybe one can change that in the Doxy settings file?

image

Anyhow, changed.

Finii:
Hmm, I do not know what Github is doing, but even a ff merge changes the commit hashes, so the tag needs to be adapted:

image

Finii:
@alt-graph Is the mirroring automatic, or shall I push to gitlab? Not that Tim builds with the old lib :grimacing:

Finii:
Mirroring is claimed but not set up??

image

Finii:
Did it just manually ... any reason this is not automated?

image

alt-graph:
> Did it just manually ... any reason this is not automated?
> 
> image

We don't have the paid version of Gitlab, so we cannot automate it from there. We could do it from Github, I guess.

PR #24, opened 2022-11-10
Add a text file on how to create a vcpkg port

alt-graph:
**[why]**
Because I usually forget how it works it between our releases.


Finii (approved):
Looks good, albeit I can not check the contents ;-)

But would it make more sense to create an `*.md` instead?

Finii:
> But would it make more sense to create an `*.md` instead?

A markdown can be easily included into our Wiki (if we ever want to fill it, but someone must have activated it for this repo)
https://github.com/gul-cpp/gul14/wiki

PR #25, opened 2022-11-16
Draft: Add Ryu to GUL

Finii:
This is a very old pull request, that we closed years ago. (`AuthorDate: Wed Aug 14 13:24:35 2019 +0200`)
The motivation to pull this out of the archives can be found further down in the comments.

Note that the code is rather old and should be revised if we decide to trudge on with this. There is a memory leak detected.
Ryu's original repo is (was) here: https://github.com/ulfjack/ryu

Lets see how it went:

----

**[why]**
The question came up whether we want to have a Grisu implementation in GUL. The more logical solution would be the `{fmt}` library. But some experimentation showed that the Grisu implementation there is hard to access directly.

If just a Grisu-like functionality is needed Ryu might be even better. And it has a direct and easy interface.

**[how]**
Forked Ryu on Github and added meson and catch2 support: https://github.com/Finii/ryu

In that repository you can also find a very basic example to play around (example.cc) done by me.

Used that to incorporate it into GUL.

**[note]**
If we would decide to really pull this (which I do not expect) we need to add copyright information stuff, add Doxygen, ....

_Edit: Add AuthorDate_


Finii:
There was an anonymous GUL user on Mattermost with this example (*)

```c++
cat(23.43, ", ", 1e+40, ", ", 1e-40)

Result:
"23.430000, 10000000000000000303786028427003666890752.000000, 0.000000"
```

which shows that `cat()` is rather useless for outputting floating point numbers.
The precision is ridiculously high for big values and unusable for small values.

This is what all the dragons do better, with `ryu` being the newest one (at least in 2019). Before that I used Grisu.
At least in 2019 `catch2` also used `ryu`, iirc.

So the question / problem is ... do we want to add some fast floating-point output stuff?

_(*) Of course the anonymous user took the example from cppreference `to_string()`, but shows how useless the output is in the general case._

Finii:
This is too old, and (astonishingly) new algorithm are STILL developed in this field.

PR #26, opened 2022-12-05
Update DOOCS release information

Finii:


Finii:
I also, manually added the tag

PR #29, opened 2022-12-22
Remove string_view::to_string()

alt-graph:
This PR removes `gul14::string_view::to_string()`. C++17's `std::string_view` does not have that member function, so we should not encourage our users to depend on it.

See: [cppreference on std::string_view](https://en.cppreference.com/w/cpp/string/basic_string_view)

### DIFF
--- a/include/gul14/string_view.h
+++ b/include/gul14/string_view.h
@@ -7,7 +7,7 @@
  * \copyright
  * Copyright Marshall Clow 2012-2015
  * Copyright Beman Dawes 2015
- * Copyright \ref contributors 2018-2020 (modifications for GUL)
+ * Copyright \ref contributors 2018-2022 (modifications for GUL)
  *
  * Distributed under the Boost Software License, Version 1.0. (See \ref license_boost_1_0
  * or http://www.boost.org/LICENSE_1_0.txt)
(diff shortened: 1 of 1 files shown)


Finii:
Was the reason to have `to_string()` not the absence of a `std::string` constructor from `string_view` (which of course the 'real' `string_view` has)? Or conversion operator, or something?

Finii:
I have the hunch that we discussed this or other custom-additions to the `std::` implementation when we added `string_view`.
Can not find that in Gitlab, maybe it has been in Stash, but that repo is deleted (instead of read-only-ed).

Like these:
image


In principle I believe removing would be fine, but we should not shoot ourselves in the foot ;)

alt-graph:
> I have the hunch that we discussed this or other custom-additions to the `std::` implementation when we added `string_view`. Can not find that in Gitlab, maybe it has been in Stash, but that repo is deleted (instead of read-only-ed).

I think we discussed the operators. But in any case `to_string()` is obsolete as a member function because `string_view` has an explicit conversion to `std::string`:
image

Finii (approved):
Hope not too many uses already in dependents.

alt-graph:
> Hope not too many uses already in dependents.

I agree. I just saw Olaf use it in code review and got curious, hence the PR...

PR #36, opened 2023-03-16
Fix several warnings from -Wshadow and -Wconversion

alt-graph:
... more fixes for warnings. I am not quite sure about some of the casts around the accessor functions.


Finii (approved):
Looks good.

I just would have called `acc` `accu`. Maybe.

PR #38, opened 2023-03-21
Add gul14::expected as a backport of std::expected from C++23

### LINKED ISSUE #27: Feature request: expected
It would be great to have a backport of [std::expected](https://en.cppreference.com/w/cpp/utility/expected) from C++23 in GUL14.

alt-graph:
This adds an implementation of the `expected` class template, (roughly) as standardized in C++23. The implementation is by Sy Brandt: https://github.com/TartanLlama/expected

I made some small adjustments like removing lots of #ifdefs that are not relevant if we assume a C++14 compiler. Also, I made the `bad_expected_access` exception class template more compliant to what the standard (or at least cppreference) says it should look like.

This also bumps the GUL14 version to 2.8.0 – at least I think so. Do we have any handling for the "patch" number that we introduced with 2.7.1?

Closes #27.


alt-graph:
Rebased on `main`. Still needs a review to be merged. :)

Finii (changes requested):
Two small things

* https://github.com/gul-cpp/gul14/blob/main/debian/copyright.in needs to be expanded I guess
* It is not documented which version you add. Maybe add to the commit message
  * Original repo link
  * Version of that repo's release that we actually added

alt-graph:
>     * https://github.com/gul-cpp/gul14/blob/main/debian/copyright.in needs to be expanded I guess
> 
>     * It is not documented which version you add. Maybe add to the commit message
>       
>       * Original repo link
>       * Version of that repo's release that we actually added

Good points, thank you! I have added the CC0 license to copyright.in and added the link to the original Github repo and version information to expected.h itself and to the first commit message.

PR #39, opened 2023-03-29
Add missing cstdint include

alt-graph:
This is a bugfix for the missing include of a standard library header as discovered by MSK's Jenkins build. Please review this with priority (but also note that another pull request is still in the queue). :)

[why]
On the openSUSE Tumbleweed Linux distro, compilation of to_number.h fails because of a missing declaration of uint64_t.

[how]
Include cstdint. Also, use the fully namespace-qualified form std::uint64_t as mandated by the standard.

### DIFF
--- a/include/gul14/to_number.h
+++ b/include/gul14/to_number.h
@@ -4,7 +4,7 @@
  * \authors \ref contributors
  * \date    Created on 19 July 2019
  *
- * \copyright Copyright 2019-2020 Deutsches Elektronen-Synchrotron (DESY), Hamburg
+ * \copyright Copyright 2019-2023 Deutsches Elektronen-Synchrotron (DESY), Hamburg
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU Lesser General Public License as published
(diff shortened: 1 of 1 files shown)


soerengrunewald (approved):
Maybe we should extend our CI with a more recent distro (e.g Ubuntu 22.04, AlmaLinux9, ArchLinux).

ohensler:
Looks good to me

alt-graph:
> Maybe we should extend our CI with a more recent distro (e.g Ubuntu 22.04, AlmaLinux9, ArchLinux).

Our Linux tests on Github are already using "Ubuntu-latest", which AFAIK currently translates to 22.04...

Finii:
> Maybe we should extend our CI with a more recent distro (e.g Ubuntu 22.04, AlmaLinux9, ArchLinux).

Well, we do run on 22.04, no?
https://github.com/actions/runner-images
We specify `ubuntu-latest` which is indeed not always quite up to date and usually there is a more recent release available by specifying specifically, but not at the moment.

Using anything not 'windows', 'mac-os', 'ubuntu' needs more work, though ;-)

What I did notice that we do not build on Windows; I believe we did that on Gitlab on 'my personal gitlab runner', but obviously that has not been 'ported'. I believe Lars did set up the CI here?

Finii:
Previously we had windows, see here, running on my 'mini desktop on window sill of old office' :grimacing: 

image

For some reason I can not access the runner definitions anymore for GUL on gitlab? :thinking:

PR #44, opened 2023-06-09
Doc/add doxygen modules for headers

### LINKED ISSUE #42: Documentation does not list header files
It would be helpful if the documentation of individual classes or functions could list the header in which they are declared. We recommend including `gul14/gul.h`, but for people who want to keep their compile time to a minimum this would be helpful.

alt-graph:
This PR groups the Doxygen documentation into one page per header file, see screenshot:
image

This makes looking up individual functions or classes much nicer because you land on a page documenting just the corresponding header, not the entire library.

Closes #42.


Finii (approved):
Yes, lets try this. I fnd the Doxygen docs rather hard to use. Maybe this will make it better (but I doubt that).

Issue #45, opened 2023-06-13
Compilation errors on Visual Studio


alt-graph:
Compilation fails on Windows. We need a CI setup for that. :)
```
"cl" "-Itests\libgul-test.exe.p" "-Itests" "-I..\..\..\tests" "-Iinclude" "-I..\..\..\include" "-Isrc" "-I..\..\..\src" "/MDd" "/nologo" "/showIncludes" "/utf-8" "/W4" "/EHsc" "/std:c++14" "/permissive-" "/Od" "/Zi" "/Fdtests\libgul-test.exe.p\test_variant.cc.pdb" /Fotests/libgul-test.exe.p/test_variant.cc.obj "/c" ../../../tests/test_variant.cc
C:\Data\Develop\src\doocs\library\common\gul14\include\gul14/variant.h(716): error C2653: 'lib': is not a class or namespace name
C:\Data\Develop\src\doocs\library\common\gul14\include\gul14/variant.h(716): error C4430: missing type specifier - int assumed. Note: C++ does not support default-int
C:\Data\Develop\src\doocs\library\common\gul14\include\gul14/variant.h(716): error C2988: unrecognizable template declaration/definition
C:\Data\Develop\src\doocs\library\common\gul14\include\gul14/variant.h(716): error C2143: syntax error: missing ';' before '<'
C:\Data\Develop\src\doocs\library\common\gul14\include\gul14/variant.h(716): error C7525: inline variables require at least '/std:c++17'
C:\Data\Develop\src\doocs\library\common\gul14\include\gul14/variant.h(716): error C2238: unexpected token(s) preceding ';'
C:\Data\Develop\src\doocs\library\common\gul14\include\gul14/variant.h(716): error C2059: syntax error: '<'
C:\Data\Develop\src\doocs\library\common\gul14\include\gul14/variant.h(717): error C2334: unexpected token(s) preceding '{'; skipping apparent function body
C:\Data\Develop\src\doocs\library\common\gul14\include\gul14/variant.h(785): error C2760: syntax error: unexpected token '{', expected 'expression'
C:\Data\Develop\src\doocs\library\common\gul14\include\gul14/variant.h(751): error C7510: 'push_back_t': use of dependent template name must be prefixed with 'template'
[...]

PR #47, opened 2023-06-13
Add CI setup for Windows builds on Github

alt-graph:
This adds a configuration to build GUL14 on a Windows CI machine with Github actions. The trickiest part is getting the equivalent of a "Developer command prompt" which knows where to find the Visual Studio toolchain. We use the msdevshell action for that.


alt-graph:
So far, it looks as if this actually worked. The failing CI for Windows shows the error from #45.

.github/workflows/build-and-run-tests-windows.yml
```diff
@@ -0,0 +1,50 @@
...
+    - name: Setup Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.x'
+
+    - name: Install Python Dependencies
+      run: pip install meson ninja
+
+    - name: Prepare MSVC
+      uses: bus1/cabuild/action/msdevshell@v1
```
Finii, line 29: This in fact duplicates our `make_vs`. We should either drop that helper script (which I wont), or use our script to check if that still works.

.github/workflows/ci.yml
```diff
@@ -13,3 +13,8 @@ jobs:
     uses: ./.github/workflows/build-and-run-tests-posix.yml
     with:
       platform: macos-latest
+
+  build-and-run-tests-on-windows:
+    uses: ./.github/workflows/build-and-run-tests-windows.yml
+    with:
+      platform: windows-latest
```
Finii, line 20: I'm completely opposed to our current setup with `ci.yml`. What is the purpose of it AT ALL?
Yes, you need this kind of workaround on Gitlab, but not on Github.

This is a great opportunity to remove it completely.

tools/make_vs.bat
```diff
@@ -106,20 +106,33 @@ ninja -C "%REPO_ROOT%%FOLDER%" %TARGET%
...
-        set VC_YEAR=2019\
-    ) ELSE IF EXIST "%VC_PREFIX%2017" (
-        set VC_YEAR=2017\
-    ) ELSE exit /B 1
+    @set VC_PREFIX=C:\Program Files\Microsoft Visual Studio\
+    @IF EXIST "%VC_PREFIX%2022" (
+        set VC_YEAR=2022\
+    ) ELSE (
+        echo "Can not find '%VC_PREFIX%2022'"
+        @set VC_PREFIX=C:\Program Files (x86^^^)\Microsoft Visual Studio\
```
alt-graph, line 114: @Finii:
```suggestion
        @set VC_PREFIX=C:\Program Files (x86)\Microsoft Visual Studio\
```
... or are these characters in some way intentional?

Finii (changes requested):
I pushed two commits, please have a look.
This is not ready with my two commits.

And apart from removing `ci.yml` I believe we should not run it at every push, compare my added new workflow.

tools/make_vs.bat
Finii: Without the escape the closing parens will close the indicated opening one

image

.github/workflows/build-and-run-tests-windows.yml
```diff
@@ -18,9 +18,9 @@ jobs:
       uses: actions/checkout@v3
 
     - name: Setup Python
-      uses: actions/setup-python@v2
+      uses: actions/setup-python@v4
       with:
-        python-version: '3.x'
+        python-version: '3.10'
```
Finii, line 23: Why 3.10 and not 3.x which is 3.11 at the moment?

tools/make_vs.bat
alt-graph: :+1: Here's me hating batch files.

.github/workflows/ci.yml
Finii: > Which workaround do you mean?

Why do we have a `ci.yml` at all?

Usually you have such a setup at Gitlab, where you can not have multiple 'root' workflows that all run in parallel - you need one dummy workflow that calls another workflow file multiple times.
With Github that is not needed.

alt-graph: I believe I set it up this way to easily call the same "Posix" workflow on two different platforms. Adding a somewhat different Windows workflow came naturally. But if you have a better solution, go ahead.

.github/workflows/build-and-run-tests-windows.yml
alt-graph: ... because I did not understand the [x-ranges](https://github.com/npm/node-semver#x-ranges-12x-1x-12-) stuff. Feel free to change back to 3.x.

.github/workflows/build-and-run-tests-windows.yml
Finii: I think the `make_vs test` does only run the tests and no asan etc

yes...
```
%MESON% --buildtype=%BUILDTYPE% "%REPO_ROOT%%FOLDER%" "%REPO_ROOT:~0,-1%"
ninja -C "%REPO_ROOT%%FOLDER%" %TARGET%
```

Do we need an asan build on windows? Do we expect it finds address issues we do not find on other platforms?

The current state of this PR has two concurrent windows ci runs:
`windows.yml` just runs the tests, is independent of `ci.yml`, triggers only when it is needed
`build-and-run-tests-windows.yml` is called from `ci.yml` (which triggers on any push) and runs asan.

Of course one can combine somehow the two workflows, pick what is best of each.

>  But if you have a better solution, go ahead

Well, I already put commits in the PR. There just needs to be a decision what we want. The code is there and we can mix.

Further I find the names are too long. Both of the files themselves as well as the workflow names. It even gets gut off:

image

image

.github/workflows/ci.yml
Finii: > I believe I set it up this way to easily call the same "Posix" workflow on two different platforms

That can also be done within one workflow file :-)

.github/workflows/build-and-run-tests-windows.yml
alt-graph: > Do we need an asan build on windows? Do we expect it finds address issues we do not find on other platforms?

I would argue we don't exactly *need* any CI at all, but every different platform/sanitizer/compiler helps a bit. So maybe we should support ASAN from our batch file at some point. Maybe not in this PR.

.github/workflows/ci.yml
alt-graph: > That can also be done within one workflow file :-)

Then please. go ahead. I have no desire at all to mess with the CI setup, and you usually see this in the quality of the result... :)

.github/workflows/build-and-run-tests-windows.yml
alt-graph: > Of course one can combine somehow the two workflows, pick what is best of each.

`windows.yml` works for me, but then we need an equivalent rewrite of `ci.yml` and the Posix file.

Finii:
Ich hab das mal aufgeraeumt. Schau mal ob dir das so gefaellt.

Dazu ein force-push um den Merge-Commit zu entfernen.

Interessanterweise macht der Windows Workflow variierende komische Fehler bei mir, als ob die Runtime begrenzt waere o.ae. ??!

Finii:
Achja, die `test*` workflows laufen jetzt nur noch bei PR. Also als klassischer 'ist dieser PR pull ready' Test.
Nicht als 'trigger a release' workflow.

Da wuerde man einen Workflow machen der auf push _only on main_ triggert.

.github/workflows/testposix.yml
```diff
@@ -0,0 +1,75 @@
...
+
+      - name: Build and run release tests
+        run: ./meson.py test -C build.release
+
+      - uses: actions/upload-artifact@v3
+        if: ${{ always() }}
+        with:
+          name: build-dirs-${{ inputs.platform }}
+          path: |
+            build.release/
```
Finii, line 74:
```suggestion
```

This has to be removed of course.

Finii:
```
libgul-test.exe is a Catch v2.13.7 host application.

Run with -? for options



-------------------------------------------------------------------------------

Compare unescaped strings

-------------------------------------------------------------------------------

../tests/test_escape.cc(44)

...............................................................................



../tests/test_escape.cc(44): FAILED:

due to a fatal error condition:

  SIGSEGV - Segmentation violation signal



===============================================================================

test cases:  43 |  42 passed | 1 failed

assertions: 338 | 337 passed | 1 failed
```

:thinking:

Finii:
Disabled asan for the time being, maybe someone else has an idea...
This seems ready to rumble from my side.

Finii:
See run for example here https://github.com/Finii/gul14/actions/runs/5259630853/jobs/9505458266

alt-graph:
> Compare unescaped strings
> -------------------------------------------------------------------------------
> ../tests/test_escape.cc(44)
> ...............................................................................
> ../tests/test_escape.cc(44): FAILED:
> 
> due to a fatal error condition:
>   SIGSEGV - Segmentation violation signal

Whoa... should we turn this into an issue? Or was this something weird like a broken build?

.github/workflows/testposix.yml
```diff
@@ -0,0 +1,75 @@
...
+
+      - name: Create symbolic link for Meson
+        run: |
+          ln -s meson-0.63.0/meson.py .
+          ls -la
+
+      - name: Configure build directories
+        run: |
+          ./meson.py --buildtype=release build.release
+          ./meson.py --buildtype=debug build.asan -Db_sanitize=address
```
alt-graph, line 61: While we're at it, maybe we should add the "setup" argument. Newer versions will complain if it is missing.
```suggestion
          ./meson.py setup --buildtype=debug build.asan -Db_sanitize=address
```

.github/workflows/testwindows.yml
```diff
@@ -0,0 +1,62 @@
...
+      - name: Intro
+        run: echo "Branch ${{ github.ref }} on repository ${{ github.repository }}"
+
+      - name: Check out repository code
+        uses: actions/checkout@v3
+
+      - name: Setup Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.10'
```
alt-graph, line 28: 3.10 again, not 3.x?

alt-graph (commented):
Looks excellent! Since this is technically still my own PR, I cannot approve, though. :+1:

Finii:
> Whoa... should we turn this into an issue? Or was this something weird like a broken build?

I had that (and other strange errors) only with the asan run on CI. If I did the same asan runs locally on the gitlab windows runner machine there is no issue at all. I guess that is some other problem, nothing related to `escape` but to asan.

.github/workflows/testwindows.yml
Finii: Well, you wrote this:

> Luckily, setup-python@v4 is available by now. We also specify a Python version of 3.10 to be on the safe side for recent Meson versions.

and I did not want to 'overrule' that. I guess it does not make much difference. We would also need to update `setup-python` if they release a new version, so there is always something to do with updates.

So in fact, if we prescribe `@4` it could also make sense to prescribe a Python version and not just take the latest one.

Finii (approved):

Finii:
I just merge this now and trigger the workflow(s) manually, links will be here:

PR #48, opened 2023-06-13
Fix compilation errors in gul14::variant when compiled with C++17

alt-graph:
As reported by Sören, when compiled with a C++17 compiler, `gul14::variant` causes a compilation error related to `invoke_result_t`. This error does not occur when compiled under C++14.

The bug seems to be subtle. Instead of fixing it directly, we can take advantage of the fact that the various invoke functions and types
```
invoke
invoke_result
invoke_result_t
is_invocable
is_invocable_r
```
are already available in namespace std on a C++17 compiler. Hence, we add a conditional compilation section that simply uses the following declarations on C++17 instead of the home-brew ones:
```cpp
using std::invoke;
using std::invoke_result;
using std::invoke_result_t;
using std::is_invocable;
using std::is_invocable_r;
```

### DIFF
--- a/include/gul14/variant.h
+++ b/include/gul14/variant.h
@@ -158,6 +158,18 @@ struct is_reference_wrapper<std::reference_wrapper<T>>
     : std::true_type {};
 
 
+// invoke, invoke_result, invoke_result_t, is_invocable, is_invocable_r
+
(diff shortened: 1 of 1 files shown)


Finii (approved):
Very elegant!

PR #49, opened 2023-06-13
Bump version to 2.9.1 and add info on DOOCS release 23.6.0

alt-graph:
We have fixed a few compilation errors on Windows and when compiled under C++17 and should therefore bump the patch number.

### DIFF
--- a/data/doxygen.h
+++ b/data/doxygen.h
@@ -143,8 +143,13 @@ namespace gul14 {
  *
  * \section changelog_2_x 2.x Versions
  *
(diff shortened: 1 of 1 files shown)


alt-graph:
... I assume the .1 in 2.9.1 just comes from the Git tag, correct?

Finii:
> ... I assume the .1 in 2.9.1 just comes from the Git tag, correct?

Yes.

Finii (approved):

PR #50, opened 2023-06-14
Bugfix for gul14::visit()

alt-graph:
It seems I underestimated the import of the `variant` header. In fact, `gul14::visit()` could not be used at all (at least for some set of compilers) because of a missing namespace qualifier on one call to `invoke()`. Here's a bugfix.

### DIFF
--- a/include/gul14/variant.h
+++ b/include/gul14/variant.h
@@ -349,7 +349,6 @@ using is_nothrow_swappable =
 } // namespace detail_variant
 
 #define AUTO_REFREF_RETURN(...) { return __VA_ARGS__; }
-#define DECLTYPE_AUTO_RETURN(...) { return __VA_ARGS__; }
 
 [[noreturn]] inline void throw_bad_variant_access() {
     throw bad_variant_access{};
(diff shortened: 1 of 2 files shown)


Finii (approved):
I think I can contribute nothing here.
The form looks good.

Finii:
Manually triggered tests:

* https://github.com/gul-cpp/gul14/actions/runs/5269569606
* https://github.com/gul-cpp/gul14/actions/runs/5269567671

Finii:
That is why I want to have C++17 proper (and a matching GUL version) ;-)
We are in 2023 now. C++20 should be our bleeding edge and not 17 :grimacing: 

Maybe adding "their" test suite completely is a good approach, if that is not too much work. I imagine that can be also quite time consuming, adapting all the namespaces...

alt-graph:
> That is why I want to have C++17 proper (and a matching GUL version) ;-) We are in 2023 now. C++20 should be our bleeding edge and not 17 😬

We should discuss if we really want to invest the time to make a GUL17. My hope is that we could jump directly to GUL20 next year, but if that is not realistic, we should move to 17 ASAP.

> Maybe adding "their" test suite completely is a good approach, if that is not too much work. I imagine that can be also quite time consuming, adapting all the namespaces...

... and it would be sheer luck if anyone used Catch2 for their test suite. Rewriting that for a different framework would be painful.

Finii:
> really want to invest the time to make a GUL17

I see that only as dropping stuff that is in the Standard. For GUL20 we would also need to drop the same stuff. So no work is lost I guess, and GUL20 becomes easier. Maybe I should just create it (it = GUL17) as workflow :-D
With some luck it will survive additions to GUL14 and automatically adapt. How hard can that be :grimacing:

alt-graph:
>I see that only as dropping stuff that is in the Standard. For GUL20 we would also need to drop the same stuff. So no work is lost I guess, and GUL20 becomes easier. Maybe I should just create it (it = GUL17) as workflow :-D
> With some luck it will survive additions to GUL14 and automatically adapt. How hard can that be 😬

You forget the migration of the DOOCS libraries and servers. There is quite a chunk of work there, and it makes a difference if we do that once or twice in the next two years.

PR #51, opened 2023-06-14
CI: Really trigger on PRs

Finii:
[why]
The test workflows are not triggered on PRs.

[how]
Correct the path filter... forgot the glob.

Also add meson.build to the filter, as changes there might also break builds.

Noticed when reviewing #50


Finii:
How could that have happened?! :grimacing:

PR #52, opened 2023-06-15
meson: enable more warnings by default

soerengrunewald:
This will enable more warnings by default. As of right now this will not trigger any new warnings.

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -18,7 +18,8 @@ if git_exe.found()
     # Get full description based on git tag API version ('v0.2-22-g3d8867c-dirty')
     git_f_api = run_command(
         git_exe, '-C', meson.current_source_dir(),
-            'describe', '--tags', '--always', '--dirty', '--match', 'v*'
+            'describe', '--tags', '--always', '--dirty', '--match', 'v*',
+        check: false
     )
     if git_f_api.returncode() != 0
         git_full_api_version = 'unknown'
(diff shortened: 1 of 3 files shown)


src/meson.build
```diff
@@ -18,7 +18,8 @@ if git_exe.found()
     # Get full description based on git tag API version ('v0.2-22-g3d8867c-dirty')
     git_f_api = run_command(
         git_exe, '-C', meson.current_source_dir(),
-            'describe', '--tags', '--always', '--dirty', '--match', 'v*'
+            'describe', '--tags', '--always', '--dirty', '--match', 'v*',
+        check: true
```
alt-graph, line 22: This "check: true" will probably break the build if you have a `git` executable, but your source code is embedded in some other weird git repo (like vcpkg does it). Note that two lines later we explicitly handle this situation.

alt-graph (changes requested):
I like the warning options, but we should be extremely careful where to put the `run_command` checks. Having git or having the source code in a working copy of our own git repo should not be a prerequisite for building.

src/meson.build
soerengrunewald: Good point, I actually did not pay attention to that. Changed!

alt-graph:
Looks good. Thanks, Sören!

src/meson.build
Finii: See also

d41fd54 `meson: Allow configuration without working git`

:grimacing:

PR #53, opened 2023-06-15
Add initialization helper for std::array

soerengrunewald:
I have found this a while ago as merge request in the [dxvk](https://github.com/doitsujin/dxvk) project. It simplifies the initialization of std::array, as one does not have to specify the actual array size. Instead we let the compiler figure it out.

There something similar available with [std::experimental::make_array](https://en.cppreference.com/w/cpp/experimental/make_array).

### DIFF
--- a/include/gul14/array_util.h
+++ b/include/gul14/array_util.h
@@ -1,0 +1,31 @@
+/* SPDX-License-Identifier: GPL-2.0 */
+/*
+ * \Copyright 2022 Joshua Ashton
+ * \Copyright 2022 DESY
+ */
+
+#ifndef GUL14_ARRAY_UTIL
+#define GUL14_ARRAY_UTIL
+
+#include <array>
+#include <utility>
+
+namespace gul14 {
+
+/**
+ * This simplifies the usage of std::array, as one does not have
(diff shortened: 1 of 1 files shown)


Finii:
Tests trigger :tada: @alt-graph ;)

alt-graph:
Nice little helper! I wouldn't create another header for it, though. Maybe we can stuff it into utility.h or traits.h?
Ah yes, tests and documentation... :) 
Don't know about the copyright situation. Can we just rewrite it from scratch? Otherwise we need an entry in our copyright list.

alt-graph:
Maybe `make_array()` would also be a possible name.

Finii:
No answer from author within 4 months, so closing...


Maybe use this implementation from cppreference:

image

Finii:
> Maybe `make_array()` would also be a possible name.

image

`to_array` would be the backport.

soerengrunewald:
I played a bit around in the attempt to backport [`std::to_array`](https://en.cppreference.com/w/cpp/container/array/to_array), bit it works only with primitive types and strings. More complex types make trouble, so I will close this MR for now.

PR #54, opened 2023-06-15
Add OverloadSet and make_overload_set()

alt-graph:
This PR adds the `OverloadSet` variadic class template and a `make_overload_set()` helper function to create objects of this type. Such a type is recommended for use with `std::visit` by Bjarne Stroustrup in "A Tour of C++" where he writes that "[t]he **overloaded** class is necessary and strangely enough, not standard". Well, at least we should have it. :)

I think we should squeeze this in as patch release 2.9.2. Yes, we enhance the API, but since this is header-only stuff, users can safely ignore it.
  
# [why]
Without C++17, we do not have "if constexpr". This can make it hard to
construct a suitable visitor function for gul14::visit(). The
alternative to using "if constexpr" is constructing an overload set of
functions.
    
# [how]
Provide a variadic class template OverloadSet that inherits from an
arbitrary number of function objects (typically lambdas) and imports
their definitions of operator() into an overload set for its own
operator(). Because constructing such an OverloadSet object manually is
almost impossible without C++17's advanced template argument deduction,
we also provide a helper function make_overload_set() to take care of
that.

### DIFF
--- a/include/gul14/variant.h
+++ b/include/gul14/variant.h
@@ -1,6 +1,6 @@
 /**
  * \file    variant.h
- * \brief   Definition of the variant class template and associated functions.
+ * \brief   Definition of the variant class template and associated functions and types.
  * \authors Michael Park, (modifications for GUL14) Lars Froehlich
  *
  * \details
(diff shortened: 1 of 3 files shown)


include/gul14/variant.h
```diff
@@ -62,10 +62,84 @@ class bad_variant_access : public std::exception
...
+{
+    using Fct1::operator();
+    using OverloadSet<Fcts...>::operator();
+
+    OverloadSet(Fct1 f1, Fcts... fs)
+        : Fct1(f1), OverloadSet<Fcts...>(fs...)
+    {}
+};
+
+/// OverloadSet for a single function object.
```
soerengrunewald, line 114: I wonder if the single function object should be declared before the variadic version.

tests/test_variant.cc
```diff
@@ -165,3 +165,23 @@ TEST_CASE("variant: visit()", "[variant]")
     v = 42L;
     REQUIRE(gul14::visit(get_type_name, v) == "long");
 }
+
+TEST_CASE("make_overload_set()", "[variant]")
+{
+    auto v = gul14::variant<float, double, long>{ };
+
+    auto get_type_name = gul14::make_overload_set(
+        [](auto&&) -> std::string { return "FAIL"; },
```
soerengrunewald, line 174: Can we have also a test case for the failure case? E.g:
```c++
auto w = gul14::variant<bool, std::string>{ };
REQUIRE(gul14::visit(get_type_name, w) == "FAIL");
w = false;
REQUIRE(gul14::visit(get_type_name, w) == "FAIL");
```

soerengrunewald (approved):

tests/test_variant.cc
alt-graph: Good idea. I have added a test for the `auto&&` case.

include/gul14/variant.h
alt-graph: I think it does not matter. Won't change until a compiler complains.

PR #56, opened 2023-06-19
meson: Do not use global arguments when subproject

soerengrunewald:
When we use libgul as subproject the setup will fail with:
> ERROR: Function 'add_global_arguments' cannot be used in subprojects because there is no way to make that reliable.

Fixes: 114b594a821 (meson: Enable extra warnings by default)

### DIFF
--- a/debian/meson.build
+++ b/debian/meson.build
@@ -38,7 +38,7 @@ changelg_run = run_command(
     get_option('deb-vers-pack') ? '-tp' : '-t',
     get_option('deb-vers-tag') + '*',
     changelog_gen,
-    check: true,
+    check: false,
 )
 
 if changelg_run.returncode() != 0
--- a/meson.build
+++ b/meson.build
@@ -11,12 +11,6 @@ project('gul14', 'cpp',
     version : '2.9',
     meson_version : '>=0.48')
 
-# Enable additional warnings by default
-if host_machine.system() != 'windows'
-    # Both GCC and clang know how to handle these options
-    add_global_arguments('-Wshadow', '-Wconversion', language: [ 'cpp' ])
-endif
-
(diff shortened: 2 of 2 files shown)


alt-graph (approved):
Ouch! Yes, makes sense. :+1:

Finii (changes requested):
Well...

Unfortunately I did not notice the previous PR (the one that is fixed by this one):

image
 
I already have a problem with THAT, and the change in this PR does not make it any better.

### First issue

Why do you add `-Wshadow` as global argument? It **is already there**, right? Is this some kind of doppelt genaeht haelt besser?

https://github.com/gul-cpp/gul14/blob/114b594a821baf17638e71604a31f55b472918eb/src/meson.build#L95-L100

### Second issue

Why do you assume all compilers on all platforms apart from Windows support the options?
`if host_machine.system() != 'windows'` and `Both GCC and clang know how to handle these options` are orthogonal and have nothing to do with each other?
Better would be to check the availability on configure time, like, see code snippet above.

----

So why do you add the at all? Is that feature not working on your platform?
It does work on mine. Here I check out before the original change.

```
$ git checkout 866ee07ca14b
$ make mrproper
$ meson setup build
$ ninja -vC build
```

image

**Edit:**

Relevant commits:

b75ac9d `Revert "meson: Allow custom warning lvl with old meson"`
87b1c22 `meson: Re-enable shadow warning`
4f8054d `meson: Enable conversion warning`

The last commit is even co-authored by you and has the same commit message as the change introduced by this global stuff (?!)

Finii:
And BTW, this is **still** broken, despite Lars' intervention in the other PR:

image

Finii (approved):
🎉

PR #58, opened 2023-12-13
Bugfix/compilation on vs2019

alt-graph:
This PR fixes compilation for VS2019 installations on English-speaking Windows installations. It also adds a few tiny corrections to the README.

### DIFF
--- a/tools/make_vs.bat
+++ b/tools/make_vs.bat
@@ -106,20 +106,24 @@ ninja -C "%REPO_ROOT%%FOLDER%" %TARGET%
     @IF DEFINED VCVARS_x86 IF DEFINED VCVARS_x64 exit /B 0
 
     @REM Trying to find the newest Visual Studio at the default install path
-    @set VC_PREFIX=C:\Program Files\Microsoft Visual Studio\
+    @IF EXIST "C:\Program Files\Microsoft Visual Studio\" (
+        @set VC_PREFIX=C:\Program Files\Microsoft Visual Studio\
+    ) ELSE IF EXIST "C:\Program Files (x86)\Microsoft Visual Studio\" (
+        @set VC_PREFIX=C:\Program Files (x86^)\Microsoft Visual Studio\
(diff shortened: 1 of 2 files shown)


tools/make_vs.bat
```diff
@@ -106,20 +106,24 @@ ninja -C "%REPO_ROOT%%FOLDER%" %TARGET%
...
-            set VC_YEAR=2017\
-        ) ELSE (
-            echo "Can not find '%VC_PREFIX%2017' or '2019'"
-            exit /B 1
-        )
+        echo "Visual Studio base directory not found"
+        exit /B 1
+    )
+	
+	@IF EXIST "%VC_PREFIX%2022" (
```
Finii, line 118:
```suggestion
@IF EXIST "%VC_PREFIX%2022" (
```

or similar

tools/make_vs.bat
```diff
@@ -106,20 +106,24 @@ ninja -C "%REPO_ROOT%%FOLDER%" %TARGET%
...
+        set VC_YEAR=2022\
+    ) ELSE IF EXIST "%VC_PREFIX%2019" (
+		set VC_YEAR=2019\
+	) ELSE IF EXIST "%VC_PREFIX%2017" (
```
Finii, line 122: Hmm, again switch in indentation.
tabs versus spaces?

Finii (changes requested):

Finii:
Probably we should aim to change this to powershell at some point, my eyes glace over when looking at the code 😬

tools/make_vs.bat
alt-graph: Uh, sorry. Yes, tabs... that was an editor I do not use regularly. Force-pushed to correct the indentation.

tools/make_vs.bat
alt-graph: Fixed.

Finii (approved):
Very good documentation fixes, btw

PR #59, opened 2023-12-20
Features for determining platform endianness

alt-graph:
This PR adds an enum class `gul14::endian` that works like C++20's [std::endian](https://en.cppreference.com/w/cpp/types/endian) and two convenience functions `is_little_endian()` and `is_big_endian()`. Together, this offers a simple and portable solution to enquire the system's endianness at compile time.

The PR also separates the documentation for the bit manipulation functions from the numeric functions.

### DIFF
--- a/include/gul14/bit_manip.h
+++ b/include/gul14/bit_manip.h
@@ -33,7 +33,7 @@ namespace gul14 {
 
 /**
  * \addtogroup bit_manip_h gul14/bit_manip.h
- * \brief Bit manipulation.
+ * \brief Bit manipulation and testing, endianness.
  * @{
  */
 
@@ -56,6 +56,44 @@ using BitFunctionReturnType =
             and not std::is_same<std::decay_t<T>, bool>::value,
         std::decay_t<T>
     >;
+
+/**
+ * An enum to determine the endianness of multi-byte scalars on the current platform.
+ *
+ * In big-endian (Motorola) order, the most significant byte is stored first, followed by
+ * the other bytes in order of decreasing significance. Little-endian platforms (e.g.
+ * Intel) store the bytes in the opposite order. There are also (historical) platforms
+ * which do not conform to either of these conventions.
+ *
+ * \code{.cpp}
+ *  if constexpr (gul14::endian::native == gul14::endian::big)
+ *      std::cout << "This is a big-endian machine!\n";
+ * \endcode
+ *
+ * This is a backport of [std::endian](https://en.cppreference.com/w/cpp/types/endian)
+ * from C++20.
+ *
+ * \see is_big_endian(), is_little_endian()
+ * \since GUL version 2.10
+ */
+enum class endian
+{
+#if defined(__BYTE_ORDER__)
+    little = __ORDER_LITTLE_ENDIAN__,
+    big    = __ORDER_BIG_ENDIAN__,
+    native = __BYTE_ORDER__
+#elif defined(_MSC_VER) && !defined(__clang__)
+    little = 0,
+    big    = 1,
+    native = little
+#else
+    #error "Don't know how to determine machine endianness on this compiler"
+    // Just for Doxygen:
+    little, ///< Little-endian (e.g. Intel)
+    big,    ///< Big-endian (e.g. Motorola)
+    native  ///< Native endianness
+#endif
+};
 
 /**
  * Set a bit in an integral type.
@@ -185,6 +223,36 @@ auto constexpr inline bit_flip(T previous, unsigned bit) noexcept -> ReturnT {
 template <typename T>
 bool constexpr inline bit_test(T bits, unsigned bit) noexcept {
     return bits & bit_set<T>(bit);
+}
+
+/**
+ * Determine whether this platform uses big-endian (Motorola) order for storing multi-byte
+ * quantities in memory.
+ *
+ * In big-endian order, the most significant byte is stored first, followed by the other
+ * bytes in order of decreasing significance.
+ *
+ * \see is_little_endian(), gul14::endian
+ * \since GUL version 2.10
+ */
+constexpr bool is_big_endian()
+{
+    return endian::native == endian::big;
+}
+
+/**
+ * Determine whether this platform uses little-endian (Intel) order for storing multi-byte
+ * quantities in memory.
+ *
+ * In little-endian order, the least significant byte is stored first, followed by the
+ * other bytes in order of increasing significance.
+ *
+ * \see is_big_endian(), gul14::endian
+ * \since GUL version 2.10
+ */
+constexpr bool is_little_endian()
+{
+    return endian::native == endian::little;
 }
 
 /// @}
--- a/tests/test_bit_manip.cc
+++ b/tests/test_bit_manip.cc
@@ -4,7 +4,7 @@
  * \date   Created on 17 Oct 2019
  * \brief  Unit tests for bit_set(), bit_reset(), bit_flip(), bit_test().
  *
- * \copyright Copyright 2019 Deutsches Elektronen-Synchrotron (DESY), Hamburg
+ * \copyright Copyright 2019-2023 Deutsches Elektronen-Synchrotron (DESY), Hamburg
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU Lesser General Public License as published
@@ -74,6 +74,11 @@ TEMPLATE_TEST_CASE("Access MSB", "[bit_manip]",
 
     REQUIRE(gul14::bit_test(x, msb) == true);
     REQUIRE(gul14::bit_test(x, msb - 1) == false);
+}
+
+TEST_CASE("is_little_endian(), is_big_endian()", "[bit_manip]")
+{
+    REQUIRE(gul14::is_little_endian() != gul14::is_big_endian());
 }
 
 // vi:ts=4:sw=4:sts=4:et
--- a/tests/test_bit_manip.cc
+++ b/tests/test_bit_manip.cc
@@ -79,6 +79,24 @@ TEMPLATE_TEST_CASE("Access MSB", "[bit_manip]",
 TEST_CASE("is_little_endian(), is_big_endian()", "[bit_manip]")
 {
     REQUIRE(gul14::is_little_endian() != gul14::is_big_endian());
+
+    const std::uint32_t val = 0x01'02'03'04;
+    const char* chars = reinterpret_cast<const char*>(&val);
+
+    if (gul14::is_little_endian())
+    {
+        REQUIRE(chars[0] == 4);
+        REQUIRE(chars[1] == 3);
+        REQUIRE(chars[2] == 2);
+        REQUIRE(chars[3] == 1);
+    }
+    else if (gul14::is_big_endian())
+    {
+        REQUIRE(chars[0] == 1);
+        REQUIRE(chars[1] == 2);
+        REQUIRE(chars[2] == 3);
+        REQUIRE(chars[3] == 4);
+    }
 }
 
 // vi:ts=4:sw=4:sts=4:et
--- a/data/doxygen.h
+++ b/data/doxygen.h
@@ -42,6 +42,7 @@ namespace gul14 {
  *  - \ref statistics_utilities
  *  - \ref debugging_utilities
  *  - \ref numeric_utilities
+ *  - \ref bit_manipulation
  *  - \ref container_utilities
  *  - \ref GSL_utilities
  *  - \ref metaprogramming
@@ -142,6 +143,11 @@ namespace gul14 {
  * \page release_notes Release Notes
  *
  * \section changelog_2_x 2.x Versions
+ *
+ * \subsection v2_10_0 Version 2.10.0
+ *
+ * - Add gul14::endian, a backport of std::endian from C++20
+ * - Add gul14::is_little_endian() and gul14::is_big_endian()
  *
  * \subsection v2_9_2 Version 2.9.2
  *
@@ -537,8 +543,6 @@ namespace gul14 {
 /**
  * \page numeric_utilities Numeric Utilities
  *
- * The General Utility Library provides several functions dealing with numbers.
- *
  * <h4>Modifying Values</h4>
  *
  * abs():
@@ -570,6 +574,10 @@ namespace gul14 {
  * within_ulp():
  *     Determine if two numbers are almost equal, allowing for a difference of a given
  *     number of units-in-the-last-place (ULPs).
+ */
+
+/**
+ * \page bit_manipulation Bit Manipulation and Testing
  *
  * <h4>Working With Bits in Integral Values</h4>
  *
@@ -584,6 +592,19 @@ namespace gul14 {
  *
  * bit_test():
  *     Test a bit in an integral value.
+ *
+ * <h4>Endianness</h4>
+ *
+ * \ref gul14::endian "endian":
+ *     An enum to determine the endianness of multi-byte scalars on the current platform,
+ *     behaving like [std::endian](https://en.cppreference.com/w/cpp/types/endian) from
+ *     C++20.
+ *
+ * is_big_endian():
+ *     Determine if the current platform uses big-endian byte order.
+ *
+ * is_little_endian():
+ *     Determine if the current platform uses little-endian byte order.
  */
 
 /**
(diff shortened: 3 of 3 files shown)


soerengrunewald (approved):

include/gul14/bit_manip.h
```diff
@@ -57,6 +57,44 @@ using BitFunctionReturnType =
...
+#if defined(__BYTE_ORDER__)
+    little = __ORDER_LITTLE_ENDIAN__,
+    big    = __ORDER_BIG_ENDIAN__,
+    native = __BYTE_ORDER__
+#elif defined(_MSC_VER) && !defined(__clang__)
+    little = 0,
+    big    = 1,
+    native = little
+#else
+    #error "Don't know how to determine machine endianness on this compiler"
```
Finii, line 90: This is bad, because that means GUL does not work / compile on platforms that do not have `__BYTE_ORDER__` or `_MSC_VER_ etc, **even if the programmer does not want to determine the endianness.**

Finii (changes requested):
Let's see if I can come up with a solution, quickly.

include/gul14/bit_manip.h
alt-graph: Well, shouldn't we simply implement a solution for that platform, then?

Finii:
Some solution could be as follows.
That compiles ok, but fails only if the enum or functions are actually used.

```diff
--- a/include/gul14/bit_manip.h
+++ b/include/gul14/bit_manip.h
@@ -78,20 +78,17 @@ using BitFunctionReturnType =
  */
 enum class endian
 {
-#if defined(__BYTE_ORDER__)
-    little = __ORDER_LITTLE_ENDIAN__,
-    big    = __ORDER_BIG_ENDIAN__,
-    native = __BYTE_ORDER__
+#if defined(__BTE_ORDER__)
+    little = __ORDER_LITTLE_ENDIAN__, ///< Little-endian (e.g. Intel)
+    big    = __ORDER_BIG_ENDIAN__, ///< Big-endian (e.g. Motorola)
+    native = __BYTE_ORDER__ ///< Native endianness
 #elif defined(_MSC_VER) && !defined(__clang__)
     little = 0,
     big    = 1,
     native = little
 #else
-    #error "Don't know how to determine machine endianness on this compiler"
-    // Just for Doxygen:
-    little, ///< Little-endian (e.g. Intel)
-    big,    ///< Big-endian (e.g. Motorola)
-    native  ///< Native endianness
+    #define GUL14_BIT_MANIP_NO_ENDIANNESS
+    // Don't know how to determine machine endianness on this compiler
 #endif
 };
 
@@ -225,6 +222,8 @@ bool constexpr inline bit_test(T bits, unsigned bit) noexcept {
     return bits & bit_set<T>(bit);
 }
 
+#ifndef GUL14_BIT_MANIP_NO_ENDIANNESS
+
 /**
  * Determine whether this platform uses big-endian (Motorola) order for storing multi-byte
  * quantities in memory.
@@ -255,10 +254,18 @@ constexpr bool is_little_endian()
     return endian::native == endian::little;
 }
 
+#else
+// We can not determine the endianness, so fail compiling with undefined function
+constexpr bool is_big_endian();
+constexpr bool is_little_endian();
+#endif
+
 /// @}
 
 } // namespace gul14
 
+#undef GUL14_BIT_MANIP_NO_ENDIANNESS
+
 #endif
 
 // vi:ts=4:sw=4:et
-- 
2.25.1
```

include/gul14/bit_manip.h
Finii: I do not know. I think failing is ok, but not for all of GUL.

Self-implemented stuff looks easier than correctly done, iirc.
That's the reason these are compiler macros and not some standard library solution.

alt-graph: I think failing directly when compiling GUL14 is the best option. It directly points to the place where we need to fix something. Otherwise, we end up with conditional compilation and will already have to #ifdef out our unit tests.

tests/test_bit_manip.cc
```diff
@@ -76,4 +76,28 @@ TEMPLATE_TEST_CASE("Access MSB", "[bit_manip]",
     REQUIRE(gul14::bit_test(x, msb - 1) == false);
 }
 
+TEST_CASE("is_little_endian(), is_big_endian()", "[bit_manip]")
+{
+    REQUIRE(gul14::is_little_endian() != gul14::is_big_endian());
+
+    const std::uint32_t val = 0x01'02'03'04;
+    const void* ptr = &val;
+    const char* chars = static_cast<const char*>(ptr);
```
Finii, line 85: Imho this should be a `reinterpret_cast` to make the intention more clear.

tests/test_bit_manip.cc
```diff
@@ -76,4 +76,28 @@ TEMPLATE_TEST_CASE("Access MSB", "[bit_manip]",
...
+
+    if (gul14::is_little_endian())
+    {
+        REQUIRE(chars[0] == 4);
```
Finii, line 89: The lifetime of the char-wide number at `chars[0]` never began -- with C++17 we have `std::launder`, here it is implementation defined (but there are no implementations known to not work as expected).

Iirc there is no way to do this at all in a standard conforming way in <= C++14. Except using C instead.

(I believe :grimacing: )

tests/test_bit_manip.cc
alt-graph: No. A `reinterpret_cast` entails undefined behavior, a `static_cast` is blessed by the standard.

include/gul14/bit_manip.h
Finii: I believe **if** we add a function to GUL it should work 'everywhere' not only where we want to use it. There is some higher barrier that functions in GUL need to pass that otherwise I would just wave through.

If we want to add the endian test it must be (strictly) correct; which is not easy as I already pointed out. Have a look for example at [boost's implementation](https://github.com/boostorg/predef/blob/develop/include/boost/predef/other/endian.h) which is long and splits into multiple dedicated header files.

A tongue in cheek implementation is imho not what GUL is about. Than maybe better do it in the clientlib (again), the scope there is so much smaller; it has a very limited set of platforms etc pp.

---

Whatever, I think we can rely on Meson for the problematic cases, like so:
(untested)

`src/meson.build`:

```python
if host_machine.endian() == 'big'
    add_cpp_args += [ '-DGUL_PLATFORM_BIG_ENDIAN' ]
endif
```

And here something along the lines of

```c++
#if defined(__BYTE_ORDER__)
    little = __ORDER_LITTLE_ENDIAN__,
    big    = __ORDER_BIG_ENDIAN__,
    native = __BYTE_ORDER__
#elif (defined(_MSC_VER) && !defined(__clang__)) || defined(GUL_PLATFORM_BIG_ENDIAN)
    little = 0,
    big    = 1,
    native = little
#else
    little = 0,
    big    = 1,
    native = big
#endif
```

alt-graph: Well, libc++ does something very similar, so I would argue that we need not be any more generic than your average standard library:
```cpp
#  ifndef __BYTE_ORDER__
#    error                                                                                                             \
        "Your compiler doesn't seem to define __BYTE_ORDER__, which is required by libc++ to know the endianness of your target platform"
#  endif

#  if __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__
#    define _LIBCPP_LITTLE_ENDIAN
#  elif __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__
#    define _LIBCPP_BIG_ENDIAN
#  endif // __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__
```
See here: https://github.com/llvm/llvm-project/blob/main/libcxx/include/__config

alt-graph: > Edit: Thinking about it, why not rely purely on Meson to figure it out?

Because that makes the workings of the library totally dependent on a single build system. In my book, that would be a very unfortunate choice.

tests/test_bit_manip.cc
Finii: > `static_cast` conversion
>
> Converts between types using a combination of implicit and user-defined conversions.

Furthermore allowed ('blessed') by the standard is only `static_cast` to and from `void*`, which does look ugly

`const char* chars = static_cast<const char*>(static_cast<void*>(ptr));`

But especially you want **no** conversion to happen at all:

> `reinterpret_cast` conversion
>
> Converts between types by reinterpreting the underlying bit pattern.

And **that** is what you want, right?

The whole code is anyhow UB (or is it implementation defined behavior, which would be ok), and that makes it more explicit that we want to do something fishy here.

alt-graph: > And that is what you want, right?

Agreed, and changed to `reinterpret_cast`.

> The whole code is anyhow UB (or is it implementation defined behavior, which would be ok), and that makes it more explicit that we want to do something fishy here.

No, it is not UB, but implementation defined for the special case of casting to some kind of `char*` or `std::byte*`. That is what I initially missed – I thought that any `reinterpret_cast` was automatically UB, which is not true.

include/gul14/bit_manip.h
Finii: Otoh

a) All build systems have such a feature (well, except `make` maybe)
b) Why do the libraries need this information anyhow? Should they not just use `htons()`?

I think the desirability of endinesseses is sorted like this

* Avoid endianness whenever possible
* Use endian-neutral code
* Use existing conversion functions, such as htons() to convert between network byte-order and the native byte-order of the processor
* Detect endianness at run-time
* Detect endianness at configuration-time
* Detect endianness at compile-time

_[Old but still relevant source](https://sourceforge.net/p/predef/wiki/Endianness/)_

:thinking:

soerengrunewald: I think we should tackle this if it becomes a problem. Until then keep it as simple as possible.
And if somebody tries to use an old GCC in the attempt to use C++14/17+ this will be the least of the problems.

tests/test_bit_manip.cc
alt-graph: What we are doing is by definition *implementation defined*, so I fail to see the problem here?

include/gul14/bit_manip.h
alt-graph: I could add the GUL14_BIG_ENDIAN macro as a fallback for the hypothetic non-gcc/clang/msvc compiler. If it is not set, we can default to little-endian and our tests will fail if that happens to be wrong. Is that a viable compromise?

alt-graph: > b) Why do the libraries need this information anyhow? Should they not just use htons()?

That is beside the point.

Finii: > That is beside the point.

Is it? For me at least GUL is not a container where I dump in all stuff that I want to have in some project, but selected functionality that I would tell people to use.
Contrarily I wound never ever suggest to people to use `endian`. I would give them the list above.
This feels like adding some functions to GUL that use `void*`. Just out of place.

So I want to ask the question (that you also asked in the meeting), _**is GUL the right place for this?**_
Most probably I would not let the functions in the clientlib that use `endian` pass a review also; well, I did not look at them. Then why add a function that endorses such code in a library that has the explicit target of 'clean and modern' code?

tests/test_bit_manip.cc
Finii: Hmm, I guess accessing something (via a pointer) which lifetime has not started is an ill-formed program iirc.
It does work (because of the C legacy), and there is no way to program something like this in a conforming way prior to C++17, but anyhow.

> I fail to see the problem here?

There is no problem. But then we have code in GUL that meticulously handles memory alignment although there would be no problem without it. We make stuff noexcept for no real benefit but just because it seems correct. And here we just hand-wave over obvious standard violations.
And for which reasons? Just that one or two doocs libraries can keep their old code and not have the `endian` problem there?

The more I think about it the more I dislike having the function in GUL at all, sorry :disappointed:

alt-graph:
So this seems to boil down to the question "Do we need this?". The answer is plain and simple: **yes**. There is a reason why standard libraries and boost include tests for endianness.

Two C++ programs that want to share information across a network or across a file need to know how multi-byte scalars are represented as bytes, there is just no way around that. So which options do we have?
1. We *can* write entirely standard-conformant code using bit shifts and bit masks. Try doing that on an array of a million integers, though, and we keep the CPU utterly busy *even if we do not need to change any byte order at all*.
2. We can use a platform-dependent C function like `htonl()`. It converts a uint32 to another uint32. Then we need to use a cast to `char*` to get at the individual bytes again. Plus, we need to include a platform-dependent header. Plus, our 1-million integer array requires 1M calls to `htonl()`. *Even if we do not need to change the byte order at all.*
3. We *can* query some platform-dependent macros in the user code to determine the platform endianness and simply `memcpy()` our 1M-integer array if it matches the target endianness, or proceed to 1., 2., or something similar to change it.

3 is by far the most performant option (for a 1M array IMO really the only acceptable option) and you will therefore find it in every serialization library worth using. The main drawback is that platform-dependent includes and macros invade user code, hence the C++ comittee decided to add the [std::endian](https://en.cppreference.com/w/cpp/types/endian) enum in C++20, which finally provides a standard way to detect platform endianness. `gul14::endian` is just a backport of this, nothing more, nothing less.

@Finii: You certainly have a right to dislike `gul14::endian` and by extension `std::endian`, but it exists for a reason. Serialization code is low-level code, but as much as we dislike that, we need it.

Finii (approved):

Finii:
> We can write entirely standard-conformant code using bit shifts and bit masks. Try doing that on an array of a million integers, though, and we keep the CPU utterly busy even if we do not need to change any byte order at all.

I do not quite get that argument. IF we can not handle that code for performance reasons on platforms where no endianness changes is needed, and then we detect by the functions in this PR that we do need an endianness change, we would anyhow need all that code?
Or is the purpose here just to say at compile time: _Can not handle your platform's endianness because of performance reasons?_

Ah I see the last potion, you want fast code on 'native' endianness and so-slow-it-almost-does-not-work on differing endianness platforms.
Ok (see approve above).

How is the transfer of floating point numbers? Is the bitwise representation the same on all platforms?

Finii:
This is a very basic and small thing (as you said). All the problems of endianness are still on the lib user.
Maybe it would be more worthy of GUL to provide real solutions and not just a platform trait.

For example modern font files have all data encoded in big endian layout.
A font renderer is supposed to NOT convert all the data structures on opening the font, but all the file structure and so on are developed to use the data as is "in situ" without a renderer data structure. That imposes "the same problem" for font renderers that need to access integer values from a big endian data structure regardless of platform endianness, always on every access.
Take for example [1] that provides the thin wrapper `BigEndianValue<T>` that does the conversion on the fly on access (if needed), for example `.from_in_place_buf()`.

The solution provided in this PR is just the bare needed basic block but not a real helper for the problems developers face with the endianness. Maybe we should strive to provide real use-ready solutions for the actual problem?

[1] https://github.com/codyd51/axle/blob/paging-demo/rust_programs/ttf_renderer/src/parse_utils.rs

alt-graph:
> How is the transfer of floating point numbers? Is the bitwise representation the same on all platforms?

The interwebs say that even IEEE-754 floating point numbers are stored in endian-dependent byte order. If we have an architecture that does not use IEEE-754, all bets are off...

https://stackoverflow.com/questions/2782725/converting-float-values-from-big-endian-to-little-endian/2782742
https://stackoverflow.com/questions/35763790/endianness-for-floating-point

alt-graph:
> This is a very basic and small thing (as you said). All the problems of endianness are still on the lib user. Maybe it would be more worthy of GUL to provide real solutions and not just a platform trait.

I do not disagree, but it might be difficult to find an API that works well for many use cases. Do we want to swap bytes in-place, copy-and-swap to a buffer, append to a vector, write to a stream? The DOOCS clientlib now has a solution, but only for a subset of these use cases (copy-and-swap to buffer/append).

Finii:
> I do not disagree, but it might be difficult to find an API that works well for many use cases. Do we want to ...

Right. The aforementioned solution does not swap anything in memory, but only on fetching into a singular value variable/register. This can even be optimized on platforms that have both-endianess where the memory-access cpu instructions can specify what to fetch.

But to be honest I have no clue at all how doocs uses this. It all sounds like 'there is endianness code but it will not work anyhow' :-> (or be too slow or ...)

:+1:

alt-graph:
> It all sounds like 'there is endianness code but it will not work anyhow' :->

Well, yes. We do not have a big-endian machine on which we can compile anymore, so the code is untested. Therefore, it won't work. As simple as that. ;)

PR #63, opened 2024-02-02
Update GitHub actions and Meson version

alt-graph:
GitHub complains about outdated actions:
> Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/checkout@v3, actions/cache@v3, actions/upload-artifact@v3.

We now use the v4 versions of these actions. At the same time, we update Meson from 0.63.0 to 1.3.1.

### DIFF
--- a/.github/workflows/testposix.yml
+++ b/.github/workflows/testposix.yml
@@ -41,19 +41,19 @@ jobs:
         id: cache-meson
         uses: actions/cache@v4
         with:
-          path: meson-0.63.0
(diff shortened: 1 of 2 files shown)


Finii (approved):
Very good :tada:

I use checkout@4 already since some time in other repos, I guess you checked the other actions if the breaking changes affect us.

Seeing the Meson version everywhere it might make sense to put it into a variable, but then, ...

:+1:

Issue #64, opened 2024-02-03
Meson warning


Finii:
![image](https://github.com/gul-cpp/gul14/assets/16012374/ef6bf1df-7bc5-44cc-94bf-bb2f3cf1d1c4)

PR #66, opened 2024-02-09
meson: Fix deprecated warning

Finii:
**[why]**
Meson 1.3.99 gives these warnings:
```
NOTICE: Future-deprecated features used:
 * 0.56.0: {'meson.build_root'}
WARNING: Broken features used:
 * 1.3.0: {'str.format: Value other than strings, integers, bools, options, dictionaries and lists thereof.'}
```

**[how]**
We try to get the absolute path of a generated file. Since 1.4.0 we have a `get_path()` member function on the generated file object, but we can not use that on older Mesons.

So instead we just manually assemble the filename in the place where the file is generated and use that later on. Both warned-about functions removed.

Fixes: #64

### DIFF
--- a/tools/meson.build
+++ b/tools/meson.build
@@ -26,6 +26,6 @@ config_checker = configure_file(
     output : 'config_checker',
     configuration : check_conf,
 )
-config_checker_path = meson.current_build_dir() + '/config_checker'
+config_checker_path = meson.current_build_dir() / '/config_checker'
 
 # vi:ts=4:sw=4:sts=4:et:syn=conf
--- a/tools/meson.build
+++ b/tools/meson.build
@@ -26,5 +26,6 @@ config_checker = configure_file(
     output : 'config_checker',
     configuration : check_conf,
 )
+config_checker_path = meson.current_build_dir() + '/config_checker'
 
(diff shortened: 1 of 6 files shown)


tools/meson.build
```diff
@@ -26,5 +26,6 @@ config_checker = configure_file(
     output : 'config_checker',
     configuration : check_conf,
 )
+config_checker_path = meson.current_build_dir() + '/config_checker'
```
soerengrunewald, line 29:
```suggestion
config_checker_path = meson.current_build_dir() / 'config_checker'
```

Finii: :woman_facepalming: TNX

Finii:
```
WARNING: Project specifies a minimum meson_version '>=0.48' but uses features which were added in newer versions:
 * 0.49.0: {'/ with string arguments'}
 ```

GRRRRR

Finii: ich erhoehe einfach auf 49

soerengrunewald (approved):
I wonder if we could go as far as upgrading meson to 0.53 as we have no older platforms, right :thinking:

Issue #65, opened 2024-02-09
Clang warnings generated on Darwin


alt-graph:
```
In file included from ../src/trim.cc:24:
In file included from ../include/gul14/trim.h:28:
In file included from ../include/gul14/string_util.h:31:
../include/gul14/string_view.h:335:47: warning: operand of ? changes signedness: 'typename iterator_traits<const char *>::difference_type' (aka 'long') to 'unsigned long' [-Wsign-conversion]
        return iter == this->cend () ? npos : std::distance ( this->cbegin (), iter );
        ~~~~~~                                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../src/trim.cc:36:27: note: in instantiation of member function 'gul14::basic_string_view<char>::find_first_not_of' requested here
    const auto lpos = str.find_first_not_of(ws_chars);
                          ^
In file included from ../src/trim.cc:24:
In file included from ../include/gul14/trim.h:28:
In file included from ../include/gul14/string_util.h:31:
../include/gul14/string_view.h:351:72: warning: implicit conversion changes signedness: 'gul14::basic_string_view::size_type' (aka 'unsigned long') to 'std::reverse_iterator<const char *>::difference_type' (aka 'long') [-Wsign-conversion]
        const_reverse_iterator iter = find_not_of ( this->crbegin () + pos, this->crend (), s );
                                                                     ~ ^~~
../src/trim.cc:41:27: note: in instantiation of member function 'gul14::basic_string_view<char>::find_last_not_of' requested here
    const auto rpos = str.find_last_not_of(ws_chars);
                          ^
In file included from ../src/trim.cc:24:
In file included from ../include/gul14/trim.h:28:
In file included from ../include/gul14/string_util.h:31:
../include/gul14/string_view.h:365:27: warning: implicit conversion changes signedness: 'typename iterator_traits<reverse_iterator<const char *>>::difference_type' (aka 'long') to 'unsigned long' [-Wsign-conversion]
        return len_ - 1 - std::distance ( first, last );
                        ~ ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../include/gul14/string_view.h:352:48: note: in instantiation of function template specialization 'gul14::basic_string_view<char>::reverse_distance<std::reverse_iterator<const char *>>' requested here
        return iter == this->crend () ? npos : reverse_distance ( this->crbegin (), iter );
                                               ^
3 warnings generated.
```

[...]

PR #67, opened 2024-02-09
string_view: Fix signed mismatch warning

Finii:
**[why]**
Apple clang version 15.0.0 warns:

```
warning: implicit conversion changes signedness:
    'size_type' (aka 'unsigned long') to 'difference_type' (aka 'long')
    [-Wsign-conversion]
```

**[how]**
_Distances_ can be forward and backward and thus need a signed number.
_Size and index_ is often described by an unsigned number, to double the range and to avoid meaningless negative length.

This has historically been mixed without much thought, although in extreme cases the results were wrong. But usually we do not have strings that long.

To fix this first the private `reverse_distance()` function is removed, because it returns a unsigned number and must return a unsigned number from what it does internally and that does not reflect the concept of 'distance'.

Instead we take the forward distance and do the 'harder' calculations in the places where `reverse_distance()` has been used (2 places).

There are now 4 instances of mixing distance and size/index. Upon careful check it is impossible that the distance in these cases can become negative; so that we can remove the sign in confidence. This is expressed by `static_casting` to the `size_type`, whereupon the warnings disappear.

Fixes: #65

### DIFF
--- a/include/gul14/string_view.h
+++ b/include/gul14/string_view.h
@@ -297,7 +297,7 @@ public:
             return npos;
         const_iterator iter = std::find_first_of
                 (this->cbegin () + pos, this->cend (), s.cbegin (), s.cend (), traits::eq);
-        return iter == this->cend () ? npos : std::distance ( this->cbegin (), iter );
+        return iter == this->cend () ? npos : static_cast< size_type >(std::distance ( this->cbegin (), iter ));
     }
     constexpr size_type find_first_of(charT c, size_type pos = 0) const noexcept
     { return find_first_of(basic_string_view(&c, 1), pos); }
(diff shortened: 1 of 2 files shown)


Finii:
Sneaking in a pending action update 🙄

soerengrunewald:
> That been said, it good to me. Although I would add some extra braces 😉

But maybe it would be even better to cast the entire expression to `size_t` and make the `- 1` a `- 1ul` :thinking:

include/gul14/string_view.h
```diff
@@ -316,7 +316,7 @@ class basic_string_view {
             pos = len_ - (pos+1);
         const_reverse_iterator iter = std::find_first_of
                 ( this->crbegin () + pos, this->crend (), s.cbegin (), s.cend (), traits::eq );
-        return iter == this->crend () ? npos : reverse_distance ( this->crbegin (), iter);
```
alt-graph, line 319: We might consider putting parentheses around the right-hand term of the `? :` operator. I had to look up the precedence to figure out that this is correct. But we also might not... :)

alt-graph (approved):
Looks great, thanks!

include/gul14/string_view.h
Finii: If it is unclear we should put parens. Done.

soerengrunewald (approved):
If you are going to put the CI changes into a separate MR. I would love to see the CI change there. And i would also suggest to prefix the CI change with a "CI:", but I would not insist on it :wink:

PR #68, opened 2024-02-12
Provide backports of std::invoke, invoke_result, is_invokable and friends

alt-graph:
This PR extracts the definitions of the C++17 backports of
    
- invoke,
- invoke_result,
- invoke_result_t,
- is_invokable,
- is_invokable_r
    
from M. Park's variant header and makes them available in the public traits.h header in namespace gul14. The definitions are useful for GUL14 users in general, but we can also use them inside the library. This PR uses them to replace an ad-hoc implementation in our `expected` header, and they will be needed for the `ThreadPool` PR.

### DIFF
--- a/include/gul14/traits.h
+++ b/include/gul14/traits.h
@@ -106,7 +106,7 @@ using std::is_invocable_r;
 
 namespace detail_invoke {
 
-#define GUL14_RETURN(...) \
+#define GUL14_INVOKE_RETURN(...) \
   noexcept(noexcept(__VA_ARGS__)) -> decltype(__VA_ARGS__) { return __VA_ARGS__; }
 
 template <typename T>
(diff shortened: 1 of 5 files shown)


Finii:
For some reason the `_v` versions are missing here?

image

include/gul14/traits.h
```diff
@@ -1,29 +1,22 @@
...
- *
- * This program is distributed in the hope that it will be useful,
- * but WITHOUT ANY WARRANTY; without even the implied warranty of
- * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
- * GNU Lesser General Public License for more details.
- *
- * You should have received a copy of the GNU Lesser General Public License
- * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ * Distributed under the Boost Software License, Version 1.0.
+ * (See \ref license_boost_1_0 or http://boost.org/LICENSE_1_0.txt)
```
Finii, line 12: I'm not so sure we can or should change existing code's license (from LGPL to Boost). At least we would need the explicit consent of all authors for the change.

This needs to be also documented in `debian/copyright`
* License change
* Author change (addition)

include/gul14/traits.h
```diff
@@ -97,6 +90,212 @@ using remove_cvref_t = typename remove_cvref<T>::type;
...
+using std::invoke_result;
+using std::invoke_result_t;
+using std::is_invocable;
+using std::is_invocable_r;
+
+#else
+
+namespace detail_invoke {
+
+#define GUL14_RETURN(...) \
```
Finii, line 109: Now that it moved into a more generic header maybe the macro name is too broad and nondescript?

Otoh it lives only until line 187. But still.

Finii (changes requested):
I just skimmed the code and it seems to be just moved, so :+1: for that.
But licensing..:

alt-graph:
> For some reason the `_v` versions are missing here?

If I am not mistaken, the `_v` versions are variable templates, and that language feature is missing from C++14.

include/gul14/traits.h
alt-graph: I changed the entire header to Boost to avoid having two licenses in a single source file. All other authors are employed by DESY, so strictly speaking we would need a management decision. But since that won't happen I'd vote for "just do it", it is just a few lines of code after all.

`debian/copyright`: Yes, of course, I keep forgetting about that.

include/gul14/traits.h
alt-graph: That is true, I have renamed it to GUL14_INVOKE_RETURN.

alt-graph:
Rebased on main.

include/gul14/traits.h
Finii: Several ;-)

Who invented the `BOOST-1` tag? ;-D

image

Links
* https://wiki.debian.org/CopyrightReviewTools
* https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
* https://spdx.org/licenses/

image

\*hust\*

image

Finii:
```sh
meson build --prefix=/usr
cp build/debian/copyright debian/copyright
rm -Rf build
debmake -kk
```

Finii:
> If I am not mistaken, the `_v` versions are variable templates, and that language feature is missing from C++14.

Ah, its inline variables that is missing in C++14 that we would need, but yes. Variable templates are part of C++14.
Without inline it would fail if we use `is_invocable_v` two times with identical parameters.

Adding a commit with some very very simple tests (that I used to find a `is_invocable_v` formulation), maybe they are kind of how to use. Maybe just drop that commit.

Finii (approved):
This looks good.
Feel free to drop the meaningless tests, but then maybe they just raise the test coverage and help sanitizer runs.

Finii:
@soerengrunewald 

I believe the license in the `RPM` spec is wrong; but I could not find out how multiple licenses are specified there.
Also the version is wrong (`2.7` in RPM but `2.10` in reality) etc pp.

I would call RPMs unsupported. Maybe I should raise an error Issue.

include/gul14/traits.h
alt-graph: I have corrected BOOST-1 -> BSL-1.0. The rest seems to be OK.

alt-graph:
> This looks good. Feel free to drop the meaningless tests, but then maybe they just raise the test coverage and help sanitizer runs.

Oops, I think I just force-pushed over them. Sorry! I'll merge as-is, but maybe you can push the commit as a new PR?

Finii:
> Oops, I think I just force-pushed over them

`git push --force-with-lease origin ...` protects you from over-force-pushing.

alt-graph:
> `git push --force-with-lease origin ...` protects you from over-force-pushing.

Thanks, I did not know that. :+1:

Finii:
> > `git push --force-with-lease origin ...` protects you from over-force-pushing.
> 
> Thanks, I did not know that. 👍

Just learned that yesterday :-D

[So You Think You Know Git - FOSDEM 2024](https://youtu.be/aolI_Rz0ZqY)

PR #70, opened 2024-02-13
Feature is_invocable_v

Finii:
**[why]**
We backported `is_invocable<>` but there you have to write
`is_invocable<>::value` all the time.

Same with `is_invocable_r<>`.

C++17 in fact also has the helper templates with the `_v` suffix.

**[how]**
Just add the helper template, but leave out the `inline` part.
That can result in multiple identical symbols in the resulting
executable, but apart from the space issue this is 100% safe.

The duplication is from the symbol name and a small bit of the data
itself (a boolean). Duplication happens once for each TU the template is
used in with a specific signature.

Here an example when we use the same

    REQUIRE(gul14::is_invocable_v<void(*)()> == true);

in two TUs (i.e. two `test_*.cc` files):

    $ nm -C build/tests/libgul-test | grep invoc
    00000000003b7614 r gul14::is_invocable_v<void (*)()>
    00000000003b7d5c r gul14::is_invocable_v<void (*)()>

### DIFF
--- a/include/gul14/traits.h
+++ b/include/gul14/traits.h
@@ -281,6 +281,21 @@ template <typename F, typename... Args>
 using is_invocable = detail_invocable::is_invocable<void, F, Args...>;
 
 /**
+ * Helper variable template for is_invocable<>::value.
+ *
+ * This is a backport of C++17's
+ * [std::is_invocable_v](https://en.cppreference.com/w/cpp/types/is_invocable).
+ *
+ * Inline variables are not possible in C++14 so this might produce more
+ * symbols than needed.
+ *
+ * \since GUL version 2.11
+ */
+template <typename F, typename... Args>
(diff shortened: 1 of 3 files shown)


alt-graph (changes requested):
I have had another look at the *one definition rule* and found this on [cppreference](https://en.cppreference.com/w/cpp/language/definition):
> One and only one definition of every non-[inline](https://en.cppreference.com/w/cpp/language/inline) function or variable that is odr-used (see below) is required to appear in the entire program (including any standard and user-defined libraries). The compiler is not required to diagnose this violation, but the behavior of the program that violates it is undefined.

My understanding is that our `is_invocable_v` violates the ODR, and that there is no way to make it work prior to C++17. The fact that no diagnostic is required probably explains why it appears to work nonetheless, but I think we cannot count on that.

PR #71, opened 2024-02-13
deb: Some copyright improvements

Finii:
**[why]**
The license identifiers are not always the standard SPDX form.

See:
* https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/#license-short-name
*  https://spdx.org/licenses/


Finii:
Hmm, to be honest: I don't know.

The `use Expat` notice is there from the beginning, i.e. since 2012. There was some confusion what `MIT` would mean.
I always use the SPDX specifier only, for example also `CC0-1.0` instead of `CC0`.

Probably you are right.

PR #74, opened 2024-02-13
Include <functional> only when compiling under C++17 (was: invoke bug)

### LINKED ISSUE #73: std::invoke not declared when compiling under C++17
Probably a missing standard header...

```
++ -Isubprojects/libgul14/tests/504481b@@_ignore_me@sta -Isubprojects/libgul14/tests -I../subprojects/libgul14/tests -Isubprojects/libgul14/include -I../subprojects/libgul14/include -Isubprojects/libgul14/src -I../subprojects/libgul14/src -fdiagnostics-color=always -pipe -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wnon-virtual-dtor -Wextra -Wpedantic -std=c++17 -O3 -fPIC -DLIBGUL_API_VERSION=2.10 -DGUL_COMPILING_SHARED_LIB -Wshadow -Wconversion -MD -MQ 'subprojects/libgul14/tests/504481b@@_ignore_me@sta/meson-generated_.._incl_span.cc.o' -MF 'subprojects/libgul14/tests/504481b@@_ignore_me@sta/meson-generated_.._incl_span.cc.o.d' -o 'subprojects/libgul14/tests/504481b@@_ignore_me@sta/meson-generated_.._incl_span.cc.o' -c subprojects/libgul14/tests/incl_span.cc
In file included from ../subprojects/libgul14/include/gul14/span.h:30,
                 from subprojects/libgul14/tests/incl_span.cc:23:
../subprojects/libgul14/include/gul14/traits.h:99:12: error: 'std::invoke' has not been declared
   99 | using std::invoke;
      |            ^~~~~~
```

alt-graph:
We use the definitions of invoke() etc. from the std:: namespace if the compiler supports C++17, but we forgot to include the `<functional>` header to actually make the declaration visible.

Closes #73

### DIFF
--- a/include/gul14/traits.h
+++ b/include/gul14/traits.h
@@ -15,7 +15,11 @@
 #ifndef GUL14_TRAITS_H_
 #define GUL14_TRAITS_H_
 
+#if __cplusplus >= 201703L
+#define GUL14_USE_CPP17_FEATURES
 #include <functional> // for std::invoke on C++17 compilers
+#endif
+
 #include <type_traits>
 #include <utility>
 
@@ -95,13 +99,15 @@ using void_t = void;
 //
 // invoke & friends
 //
-#if __cplusplus >= 201703L
+#ifdef GUL14_USE_CPP17_FEATURES
 
 using std::invoke;
 using std::invoke_result;
 using std::invoke_result_t;
 using std::is_invocable;
 using std::is_invocable_r;
+
+#undef GUL14_USE_CPP17_FEATURES
 
 #else
 


soerengrunewald (approved):

Finii:
```
Scenario: Negative or zero times make sleep() not wait
       When: sleep(0ms) is executed
       Then: the elapsed time is very very small
-------------------------------------------------------------------------------
../tests/test_time_util.cc:88
...............................................................................

../tests/test_time_util.cc:90: FAILED:
  REQUIRE( toc(t0) < 0.001 )
with expansion:
  0.00130301 < 0.001
```

Ah MacOS problem will be solved by #76

include/gul14/traits.h
```diff
@@ -15,6 +15,7 @@
 #ifndef GUL14_TRAITS_H_
 #define GUL14_TRAITS_H_
 
+#include <functional> // for std::invoke on C++17 compilers
```
Finii, line 18:
```suggestion
#if __cplusplus >= 201703L
#include <functional> // for std::invoke on C++17 compilers
#endif
```

All in all I would put that into a new macro somewhere in the file top and reuse that local macro everywhere we switch between `std` and `gul14` in the file.

Finii (commented):
I would not include functional _always_ but only if need be.

Finii:
> MacOS failing?

Succeed after re-run. See other comment and Lars' other PR.

include/gul14/traits.h
alt-graph: > All in all I would put that into a new macro somewhere in the file top and reuse that local macro everywhere we switch between std and gul14 in the file.

I tried that in 4e9444d22cb6a6409260ec447dd06034d1ddc48b, but I do not like the result. It adds another macro and complicates the code just to shave a few milliseconds off of someone's build times. I'd vote for just using the first commit.

alt-graph: In fact, I'll cherry-pick the first commit into `main`. The discussion can continue here.

Finii: Is there a reason we use `std::invoke` if we are on C++17?
I believe we use `gul::string_view` on 17 even though `std::string_view` is available?

alt-graph:
I propose to close this PR. IMO it only adds complexity without any major gain.

Finii:
I still hope for GUL17, which already has a repo here.

Issue #72, opened 2024-02-13
RPM spec file broken?


Finii:
I have no clue about RPM packaging.

Excerpt of the `libgul14.spec`:

```
#
# This file is used to build RPMS for libgul14

Name:           libgul14
Version:        2.7
Release:        1%{?dist}
Summary:        General Utility Library

License:        GPLv2
URL:            https://winweb.desy.de/mcs/docs/gul/
Source0:        gul14-%{version}.tar.xz
```

* It appears that the version is wrong
* The license is questionable (see more detailed debian/copyright)

alt-graph:
The URL is wrong, too.

PR #77, opened 2024-02-16
Update rpm-spec file and documentation

soerengrunewald:
This should resolve issue #72. At the same time also updates the documentation a bit.

Fixes: #72 

_Edit: Add "Fixes" line to link to issue._

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,16 +44,19 @@ Clone this repository:
 
 Use meson to configure the build you want to have. A list of the
 [build options](#Build-configuration-switches) is given below. Build directory name can be
-what you want. If you use directories starting with 'build' they will be ignored by git.
+what you want and can even be located outside of the projecte directory. E.g. on Linux you
+can place the build directory into a tmpfs mount (aka RAM disk).
+If your build directories are inside the project root directory and starting with 'build'
+they will be ignored by git.
 Examples:
 
-        meson setup build
-        meson setup -D tests=false build_without_tests
-        meson setup --prefix=/usr --buildtype=release build_rel
-        meson setup --buildtype=release -D unity=on build_rel_unity
-        meson setup --buildtype=release -D docs=false build_nodocs_rel
-        meson setup -Dcpp_args='-Wconversion -Wshadow' --buildtype=debug builddir
-        CXX=clang++ meson setup build_clang
+        meson setup <build-dir>
+        meson setup -D tests=false <build-without-tests-dir>
+        meson setup --prefix=/usr --buildtype=release <build-rel-dir>
+        meson setup --buildtype=release --unity=on <build-rel-unity-dir>
+        meson setup --buildtype=release -D docs=false <build-nodocs-rel-dir>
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -47,21 +47,21 @@ Use meson to configure the build you want to have. A list of the
...
-        meson setup --buildtype=release -D unity=on build_rel_unity
-        meson setup --buildtype=release -D docs=false build_nodocs_rel
-        meson setup -Dcpp_args='-Wconversion -Wshadow' --buildtype=debug builddir
-        CXX=clang++ meson setup build_clang
+        meson setup <build-dir>
+        meson setup -D tests=false <build-without-tests-dir>
+        meson setup --prefix=/usr -D buildtype=release <build-rel-dir>
+        meson setup -D buildtype=release -D unity=on <build-rel-unity-dir>
+        meson setup -D buildtype=release -D docs=false <build-nodocs-rel-dir>
+        meson setup -Dcpp_args='-Wconversion -Wshadow' --buildtype=debug <build-with-extra-warnings-dir>
```
alt-graph, line 55: Hmmm... is `-D buildtype` the same as `--buildtype`? And why is is not changed in this line?

alt-graph (approved):
Thanks, Sören! Looks OK, maybe one remark to be addressed for the README.

README.md
Finii: Screenshot 2024-02-17 at 01 11 28

So here we have mentioned `-Dbuildtype=release` and `--buildtype=release`. In fact `-D buildtype=release` is **not** mentioned here.

When you type `meson setup --help` all the direct (without `-D`) options are listed.
For that reason
* I would prefer them here, where possible (`buildtype`, `unity`)
* use the `-D` only where needed
* not use `-D opt=value` _at all_ (yes it is documented, but not here)

libgul14.spec
```diff
@@ -99,6 +99,9 @@ Requires:       %{name}-devel%{?_isa} = %{version}-%{release}
 %{_libdir}/%{name}.a
 
 %changelog
+* Fri Feb 16 2024 Soeren Grunewald <soeren.grunewald@desy.de> - 2.10-1
+- New upstream release
+
```
Finii, line 104: For debian we invented a way to get rid of these manual changelog entries and stuff, but redhat brings that back? Not very nice. And this looks more like it belongs into a packaging repo, that fetched upstream and packages that.

Which is maybe the way to go anyhow, to not have the distribution stuff inside the (this) repo and have instead packager's repos. This is what I know and like from Arch. Well, probably all distribution releasers do it that way.

Finii (changes requested):
Meson commands can be unified a bit, otherwise fine (enough) ;-)

libgul14.spec
soerengrunewald: Right, ideally the packaging should be a separate repository/project.

README.md
soerengrunewald: Valid points, I added the `-D` options because I experimented with [`muon`](https://muon.build/) and there all options are used with `-D`. But we don't mention `muon`, so we should tick with meson style.  

I will change it back.

Finii (approved):

Issue #69, opened 2024-02-13
to_number() test fails on Alpine Linux 3.19


alt-graph:
By Soeren:

I tried to build the current release on Alpine 3.19 and got failures when testing:
```bash
1/1 all        FAIL            0.74s   exit status 4
>>> LD_LIBRARY_PATH=/tmp/build/libgul/src MALLOC_PERTURB_=64 ASAN_OPTIONS=halt_on_error=1:abort_on_error=1:print_summary=1 UBSAN_OPTIONS=halt_on_error=1:abort_on_error=1:print_summary=1:print_stacktrace=1 /tmp/build/libgul/tests/libgul-test
――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――― ✀  ――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
libgul-test is a Catch v2.13.7 host application.
Run with -? for options

-------------------------------------------------------------------------------
to_number(): min and subnormal floating point - long double
-------------------------------------------------------------------------------
../../../home/grunewal/git/desy.de/doocs/library/common/libgul/tests/test_to_number.cc:206
...............................................................................

../../../home/grunewal/git/desy.de/doocs/library/common/libgul/tests/test_to_number.cc:222: FAILED:
  REQUIRE( true == gul14::within_ulp(to_number<TestType>(ss.str()).value(), num, lenience) )
with expansion:
  true == false
with message:
  ss.str() := "3.36210314311209350626e-4932"

-------------------------------------------------------------------------------
to_number(): max and overflow floating point - long double
-------------------------------------------------------------------------------
../../../home/grunewal/git/desy.de/doocs/library/common/libgul/tests/test_to_number.cc:226
...............................................................................

../../../home/grunewal/git/desy.de/doocs/library/common/libgul/tests/test_to_number.cc:235: FAILED:
  REQUIRE( true == gul14::within_ulp(to_number<TestType>(numb).value(), max, lenience) )
with expansion:
  true == false

-------------------------------------------------------------------------------
to_number(): lowest and overflow floating point - long double
-------------------------------------------------------------------------------
[...]

PR #78, opened 2024-02-16
Bugfix inaccurate powl

Finii:
**[why]**
For `to_number<long double>()` to work we need a `std::powl()` that is
accurate - its errors directly influence our conversion output, usually
amplified.
In the past we just knew about Apple clang that had a problem with
`std::powl()` being off a bit, but now also Alpine has the same problem
via musl that it uses.

**[how]**
On parsing a string a possible number is expressed with scientific
notification, for example `123.4E5` which means 123.4 * (10^5).

The mantissa is converted to a normalized value as floating point
number, (1.234 in the example) and the exponent is corrected for the
shift of decimal places (5 becomes 7). So the example is transformed
to 1.234 * (10^7).

To calculate the exponent the call `pow(10, 7)` is used. Any inaccuracy
in the calculation directly influences the result. For very big (or
small) exponents musl is off by more than 1000 ULPs.

But in fact a much simpler function than `pow()` would be sufficient.
The base is always 10 and the exponent is always integer. Usually `pow()`
calculates the result with `log2()` and `exp()` under the hood.
But given the constraints we can get away with a much simpler approach.

The new function `pow10(integer)` uses a table of precalculated results
for the exponents -16 to 16. Values for the other (bigger/smaller)
exponents are calculated with the right to left binary exponentiation.
The table needed for the binary approach is reasonably small while
giving good enough accuracy. Despite the fact that several
multiplications are chained the result is usually still within 1 ULP
to the correct result, only exponents that have almost all bits set
(i.e. need a lot of multiplications (7 or more)) have an error
of 2 ULP.

As we now have an exact enough `pow()` substitude the platform specific
exceptions and fallback to `std::strtold()` (that needs allocations)
can be removed.

The code assumes that
* double is: IEEE 754 binary64 (11 bit exponent)
* long double is:
  * just the same as double or
  * IEEE 754 Intel extended precision (15 bit exponent) or the
        same representation padded (on ARM)

If long double has exponents with not exactly 11 or 15 bits the code
and/or tests can break (fail).

There is no additional correction step to get the error in ULP down;
but we never guaranteed a better conversion than 3 ULP off. That still
holds. With one test we need to allow 1 ULP error where before it was
0 ULPs.

(*) The assumption is that floating point literals are handled
correctly; rounded to the nearest representation = 0 ULP off of
the ideal result.

**[note]**
Here some examples of reported std::powl() accuracy:

Ubuntu 23.10 (GLIBC 2.38)

    powl(10, -4000) off by 0 ULP
    powl(10, -2000) off by 0 ULP
    powl(10, -1000) off by 0 ULP
    powl(10, -500) off by 0 ULP
    powl(10, -250) off by -1 ULP
    powl(10, -100) off by 1 ULP
    powl(10, -10) off by 0 ULP

Alipne 3.19.1 (musl 1.2.4)

    powl(10, -4000) off by 26 ULP
    powl(10, -2000) off by 11 ULP
    powl(10, -1000) off by 5 ULP
    powl(10, -500) off by 1 ULP
    powl(10, -250) off by 1 ULP
    powl(10, -100) off by 1 ULP
    powl(10, -10) off by 0 ULP

Sonoma 14.3 on M2 (libSystem 1336.61.1)

    powl(10, -250) off by 0 ULP
    powl(10, -100) off by 0 ULP
    powl(10, -10) off by 0 ULP

On Sonoma it seems double == long double.

Github MacOS CI machine (x86?) (libSystem 1319.0.0)

    powl(10, -4000) off by more than 1000 ULP
    powl(10, -2000) off by 830 ULP
    powl(10, -1000) off by 396 ULP
    powl(10, -500) off by 191 ULP
    powl(10, -250) off by 133 ULP
    powl(10, -100) off by 66 ULP
    powl(10, -10) off by 6 ULP

    which is unexpected as the double performance is better

    pow(10, -250) off by 0 ULP
    pow(10, -100) off by 0 ULP
    pow(10, -10) off by 0 ULP

**[note]**
Some useful links:

About floating point representations:
https://en.wikipedia.org/wiki/Double-precision_floating-point_format
https://en.wikipedia.org/wiki/Extended_precision

David M. Gay's strtod() (Expat) for example here:
https://portal.ampl.com/~dmg/netlib/fp/dtoa.c

Rick Regan's Exploring Binary blog, for example
https://www.exploringbinary.com/fast-path-decimal-to-floating-point-conversion/
https://www.exploringbinary.com/strtod-initial-decimal-to-floating-point-approximation/
https://www.exploringbinary.com/why-powers-of-ten-up-to-10-to-the-22-are-exact-as-doubles/

Fixes: #69

### DIFF
--- a/include/gul14/to_number.h
+++ b/include/gul14/to_number.h
@@ -4,7 +4,7 @@
  * \authors \ref contributors
  * \date    Created on 19 July 2019
  *
- * \copyright Copyright 2019-2023 Deutsches Elektronen-Synchrotron (DESY), Hamburg
+ * \copyright Copyright 2019-2024 Deutsches Elektronen-Synchrotron (DESY), Hamburg
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU Lesser General Public License as published
@@ -267,6 +267,20 @@ constexpr inline ParseInfNanResult<NumberType> parse_inf_nan(gul14::string_view 
 }
 
 /**
+ * Calculate 10 to the power of an integer exponent.
+ *
+ * This is used to calculate values for strings of the form "1.0En"
+ * and solves the "En" part (where n is an integer number).
+ *
+ * The result is a long double. This is faster than pow(10.0L, exponent)
+ * and even more accurate with some standard clibs.
+ *
+ * \param exponent  The exponent
+ */
+GUL_EXPORT
+long double pow10(int exponent);
+
+/**
  * Convert a string_view into a floating point number.
  * No sign is allowed at the beginning of the number (thus unsigned float).
  *
@@ -356,7 +370,7 @@ constexpr inline optional<NumberType> to_unsigned_float(gul14::string_view str) 
     if (not norm_val.has_value())
         return nullopt;
 
-    return NumberType(std::pow(CalcType(10), CalcType(exponent)) * *norm_val);
+    return detail::pow10(exponent) * *norm_val;
 }
 
 /**
@@ -522,11 +536,6 @@ constexpr inline optional<NumberType> to_number(gul14::string_view str) noexcept
         return nullopt;
 
     if (
-#ifdef __APPLE__
-        // Apple clang 8.0.0 has a bug with std::pow and long double types,
-        // where the result is off by a huge amount. Use std::strtold() here.
-        (sizeof(NumberType) > sizeof(double)) ||
-#endif
 #ifdef _MSC_VER
 #    pragma warning( push )
 #    pragma warning( disable: 4127 ) // conditional expression is constant
--- a/src/to_number.cc
+++ b/src/to_number.cc
@@ -1,0 +1,98 @@
+/**
+ * \file    to_number.cc
+ * \brief   Implementation of pow10().
+ * \authors \ref contributors
+ * \date    Created on 19 Februar 2024
+ *
+ * \copyright Copyright 2024 Deutsches Elektronen-Synchrotron (DESY), Hamburg
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published
+ * by the Free Software Foundation, either version 2.1 of the license, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+#include <cfloat>
+
+#include "gul14/to_number.h"
+
+namespace gul14 {
+
+namespace detail {
+
+long double pow10(int exponent)
+{
+    // The preprocessor macros below just allow long double with
+    //
+    // - exponent maximum 511 (typical is 308) which is usually
+    //   the case when long double == double (the code needs at least
+    //   an allowed maximum of 256 for the constants)
+    //
+    // OR
+    //
+    // - exponent maximum 8191 (typical 4932) which is usually
+    //   the case for 80 bit 'extended precision' or 'float128'
+    //   long doubles (the code needs at least an allowed maximum
+    //   of 4096 for the constants)
+
+    static_assert(LDBL_MAX_10_EXP >= 4096 or LDBL_MAX_10_EXP < 512,
+        "Unsupported long double type");
+    static_assert(LDBL_MIN_10_EXP <= -4096 or LDBL_MIN_10_EXP > -512,
+        "Unsupported long double type");
+
+    constexpr const long double direct_powers_pos[] = {
+        1E0L, 1E1L, 1E2L, 1E3L, 1E4L, 1E5L, 1E6L, 1E7L,
+        1E8L, 1E9L, 1E10L, 1E11L, 1E12L, 1E13L, 1E14L, 1E15L
+    };
+    constexpr const long double direct_powers_neg[] = {
+        1E0L, 1E-1L, 1E-2L, 1E-3L, 1E-4L, 1E-5L, 1E-6L, 1E-7L,
+        1E-8L, 1E-9L, 1E-10L, 1E-11L, 1E-12L, 1E-13L, 1E-14L, 1E-15L
+    };
+    constexpr const long double binary_powers_pos[] = {
+        1E16L, 1E32L, 1E64L, 1E128L, 1E256L,
+#if LDBL_MAX_10_EXP >= 4096
+        1E512L, 1E1024L, 1E2048L, 1E4096L,
+#endif
+        std::numeric_limits<long double>::infinity()
+    };
+    constexpr const long double binary_powers_neg[] = {
+        1E-16L, 1E-32L, 1E-64L, 1E-128L, 1E-256L,
+#if LDBL_MIN_10_EXP <= -4096
+        1E-512L, 1E-1024L, 1E-2048L, 1E-4096L,
+#endif
+        0
+    };
+    static_assert(sizeof(binary_powers_pos) == sizeof(binary_powers_neg),
+        "Precalculated arrays need to have the same size");
+    constexpr auto binary_powers_last_index =
+        sizeof(binary_powers_pos) / sizeof(*binary_powers_pos) - 1;
+    const long double* direct_powers = direct_powers_pos;
+    const long double* binary_powers = binary_powers_pos;
+
+    if (exponent < 0) {
+        exponent = -exponent;
+        direct_powers = direct_powers_neg;
+        binary_powers = binary_powers_neg;
+    }
+    auto result = direct_powers[exponent & 0xF];
+    exponent >>= 4;
+    for (int exponent_bit = 0; exponent > 0; ++exponent_bit, exponent >>= 1) {
+        if (exponent_bit == binary_powers_last_index)
+            return binary_powers[exponent_bit];
+        if (exponent & 1)
+            result *= binary_powers[exponent_bit];
+    }
+    return result;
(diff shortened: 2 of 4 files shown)


Finii:
If I remember correctly other projects have for that exact reason their own limited `pow()`-like function that is exact (not even off by 1 ULP), for example Ryu and the more modern variants. We could in principle also use that approach, but that would make it all more complicated. I guess we do not want that kind of tables.

soerengrunewald (approved):
Except some style nitpicks it looks good to me. -- I commented on the old version :unamused: 
1. `report_all` seem to be very much useless as it is right now.
2. `powl_error_ulp` and `find_order` should be static, just for good style.

Finii:
Configure time decisions and header only files do not mix (so easily).
Took me some time to realize the problem.

Finii:
Tried to address nitpicks.

Finii:
Here, btw, the shocking results for the Apple machine used for Gitlab CI, I guess a x86 machine like the ones used for Doocs...

```
/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1319.0.0)

powl(10, -4000) off by more than 1000 ULP
powl(10, -2000) off by 830 ULP
powl(10, -1000) off by 396 ULP
powl(10, -500) off by 191 ULP
powl(10, -250) off by 133 ULP
powl(10, -100) off by 66 ULP
powl(10, -10) off by 6 ULP
```

which is worse than the `double` performance, same machine:

```
/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1319.0.0)

pow(10, -250) off by 0 ULP
pow(10, -100) off by 0 ULP
pow(10, -10) off by 0 ULP
```

Finii:
Start from scratch, force push.

Finii:
We could make `pow10` part of GUL.
And maybe `error_in_ulp()`.
Both in a later PR.

Invite to discuss @alt-graph

src/to_number.cc
```diff
@@ -0,0 +1,99 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+#include <float.h>
```
soerengrunewald, line 23:
```suggestion
#include <cfloat>
```

src/to_number.cc
```diff
@@ -0,0 +1,99 @@
...
+        direct_powers = direct_powers_pos;
+        binary_powers = binary_powers_pos;
+    } else {
+        exponent = -exponent;
+        direct_powers = direct_powers_neg;
+        binary_powers = binary_powers_neg;
+    }
+    auto result = direct_powers[exponent & 0xF];
+    exponent >>= 4;
+    for(int exponent_bit = 0; exponent > 0; ++exponent_bit, exponent >>= 1) {
```
soerengrunewald, line 88:
```suggestion
    for (int exponent_bit = 0; exponent > 0; ++exponent_bit, exponent >>= 1) {
```

src/to_number.cc
```diff
@@ -0,0 +1,99 @@
...
+
+#include <float.h>
+
+#include "gul14/to_number.h"
+
+namespace gul14 {
+
+namespace detail {
+
+long double pow10(int exponent)
```
soerengrunewald, line 31: I think this can be made constexpr

tests/test_to_number.cc
```diff
@@ -22,7 +22,9 @@
 
 #include <array>
 #include <cmath>
+#include <float.h>
```
soerengrunewald, line 25:
```suggestion
#include <cfloat>
```

tests/test_to_number.cc
```diff
@@ -45,6 +47,25 @@ struct TestConfig {
     unsigned static const number_of_random_tests{ 5'000 };
 };
 
+template <typename T>
+inline int error_in_ulp(T value, T reference)
```
soerengrunewald, line 51:
```suggestion
int error_in_ulp(T value, T reference)
```

soerengrunewald: Tests?

tests/test_to_number.cc
```diff
@@ -420,4 +431,107 @@ TEMPLATE_TEST_CASE("to_number(): Floating-point types, constexpr", "[to_number]"
 }
 */
 
+TEST_CASE("test pow10()", "[to_number]")
```
soerengrunewald, line 434: A lot of repetition, table might be better, but readability will probably suffer. So lets keep it as is.

src/to_number.cc
```diff
@@ -0,0 +1,99 @@
...
+{
+    // The preprocessor macros below just allow long double with
+    //
```
alt-graph, line 34: Why the empty comment line?

tests/test_to_number.cc
```diff
@@ -363,7 +374,7 @@ TEMPLATE_TEST_CASE("to_number(): Convert integers with many leading zeros", "[to
 TEMPLATE_TEST_CASE("to_number(): Convert floating-point values with many leading and trailing zeros", "[to_number]",
     float, double, long double)
 {
-    auto const lenience = 0;
+    auto const lenience = 1;
```
soerengrunewald, line 377: Ok, zero lenience makes only sense for leading zero but not after the dot, right?

src/to_number.cc
```diff
@@ -0,0 +1,99 @@
...
+    // an allowed maximum of 256 for the constants)
+    //
+    // OR
+    //
+    // exponent maximum 8191 (typical 4932) which is usually
+    // the case for 80 bit 'extended precision' or 'float128'
+    // long doubles (the code needs at least an allowed maximum
+    // of 4096 for the constants)
+
+    static_assert(LDBL_MAX_10_EXP >= 4096 or LDBL_MAX_10_EXP < 512,
```
alt-graph, line 46: We could use https://en.cppreference.com/w/cpp/types/numeric_limits/max_exponent10 instead of the macros, I guess. That would in theory also allow us to make pow10 a template that can be instantiated for float, double, long double. Don't know if that makes sense, so feel free to disregard this comment.

soerengrunewald (approved):
Looks good in general. Just some minor style thingies...

src/to_number.cc
```diff
@@ -0,0 +1,99 @@
...
+        binary_powers = binary_powers_pos;
+    } else {
+        exponent = -exponent;
+        direct_powers = direct_powers_neg;
+        binary_powers = binary_powers_neg;
+    }
+    auto result = direct_powers[exponent & 0xF];
+    exponent >>= 4;
+    for(int exponent_bit = 0; exponent > 0; ++exponent_bit, exponent >>= 1) {
+        if (exponent_bit == sizeof(binary_powers_pos) / sizeof(*binary_powers_pos) - 1)
```
alt-graph, line 89: Maybe we could create a constant here, something like:
```cpp
constexpr auto num_binary_powers = sizeof(binary_powers_pos) / sizeof(*binary_powers_pos);
```
Just to make this a little easier to follow.

tests/test_to_number.cc
alt-graph, line 25: I guess this should be `<cfloat>`.

include/gul14/to_number.h
```diff
@@ -266,6 +266,21 @@ constexpr inline ParseInfNanResult<NumberType> parse_inf_nan(gul14::string_view
...
+ * Calculate 10 to the power of an integer exponent.
+ *
+ * This is used to calculate values for strings of the form "1.0En"
+ * and solves the "En" part (where n is an integer number).
+ *
+ * The result is a long double. This is faster than pow(10.0L, exponent)
+ * and even more accurate with some standard clibs.
+ *
+ * /param exponent The exponent
+ * /returns the resulting number
```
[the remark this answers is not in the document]
Finii: Duh! Tnx

src/to_number.cc
Finii: This separates the text from a list, changed it like this to make it more clear:

```
    // The preprocessor macros below just allow long double with
    //
    // - exponent maximum 511 (typical is 308) which is usually
    //   the case when long double == double (the code needs at least
    //   an allowed maximum of 256 for the constants)
    //
    // OR
    //
    // - exponent maximum 8191 (typical 4932) which is usually
    //   the case for 80 bit 'extended precision' or 'float128'
    //   long doubles (the code needs at least an allowed maximum
    //   of 4096 for the constants)
```

src/to_number.cc
Finii: As `constexpr` implies `inline` this would not work for the exported function. It can only be exported _or_ inline (i.e. constexpr), not both.

At least I believe that.

We could make it inline of course, but that would instantiate it multiple times later on, and with the data tables I believe that is not a so good idea. I'm not really sure about the concrete downsides.
But the function will not be called in a constexpr context from within to_number, and so constexpr does not gain us anything. Could be different if we pull it out of the detail namespace.

Leaving this as is for the time being.
This is related to the question if we want this as template, below.

src/to_number.cc
```diff
@@ -0,0 +1,99 @@
...
+#endif
+        std::numeric_limits<long double>::infinity()
+    };
+    static const long double binary_powers_neg[] = {
+        1E-16L, 1E-32L, 1E-64L, 1E-128L, 1E-256L,
+#if LDBL_MIN_10_EXP <= -4096
+        1E-512L, 1E-1024L, 1E-2048L, 1E-4096L,
+#endif
+        0
+    };
```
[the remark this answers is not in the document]
Finii: done

tests/test_to_number.cc
Finii: Waaah !!!

I did implement a test and **found a bug** :grimacing:

tests/test_to_number.cc
Finii: improved

PR #82, opened 2024-02-29
to_number: Relax pow10() test

Finii:
**[why]**
On ARM (Raspberry 3 for example):

Small numbers are 2 ULP off of the ideal value but we accept only 1 ULP - albeit we say in the comment we would allow 2 for very small numbers.

These are not yet denormal values.

The problem happens if the highest exponent bit is set (i.e. bit 12 with value 4096) and there are more than 5 multiplications needed.

Fixes: #81


alt-graph (approved):
:+1:

Issue #83, opened 2024-03-01
Dynamically link unit tests


alt-graph:
In #62, we had to start linking our unit tests statically. The reason (as far as we understand it) is the following:
- We had to add `threads` as a dependency because the main library can now actually spawn `std::thread`s.
- That lead to linker failures in building the tests, apparently because the test build picked up an outdated GUL14 library from the system.

This is actually reproducible: When linking against `libgul_dep` instead of `libgul_static_dep`, we get this:
```
g++  -o tests/libgul-test tests/libgul-test.p/test_backports.cc.o tests/libgul-test.p/test_bit_manip.cc.o tests/libgul-test.p/test_case_ascii.cc.o tests/libgul-test.p/test_cat.cc.o tests/libgul-test.p/test_escape.cc.o tests/libgul-test.p/test_expected.cc.o tests/libgul-test.p/test_finalizer.cc.o tests/libgul-test.p/test_gcd_lcm.cc.o tests/libgul-test.p/test_hexdump.cc.o tests/libgul-test.p/test_join_split.cc.o tests/libgul-test.p/test_main.cc.o tests/libgul-test.p/test_num_util.cc.o tests/libgul-test.p/test_optional.cc.o tests/libgul-test.p/test_replace.cc.o tests/libgul-test.p/test_SlidingBuffer.cc.o tests/libgul-test.p/test_SmallVector.cc.o tests/libgul-test.p/test_statistics.cc.o tests/libgul-test.p/test_string_util.cc.o tests/libgul-test.p/test_substring_checks.cc.o tests/libgul-test.p/test_ThreadPool.cc.o tests/libgul-test.p/test_time_util.cc.o tests/libgul-test.p/test_to_number.cc.o tests/libgul-test.p/test_tokenize.cc.o tests/libgul-test.p/test_Trigger.cc.o tests/libgul-test.p/test_trim.cc.o tests/libgul-test.p/test_type_name.cc.o tests/libgul-test.p/test_variant.cc.o -fsanitize=address -Wl,--as-needed -Wl,--no-undefined '-Wl,-rpath,$ORIGIN/../src' -Wl,-rpath-link,/home/lfroehli/doocs/library/common/gul14/builddir/src -Wl,--start-group src/libgul14.so.2.10.0 -Wl,--end-group -pthread
[...]

PR #84, opened 2024-03-06
ThreadPool: Fixup: Export missing symbols

Finii:
**[why]**
Some symbols (functions) are not exported to the shared library. But they are accessed by templated functions, so the call to them is instantiated by user code and the user needs to have access to the functions. That is even then the case when private member functions are involved.

**[how]**
Add export specifier to needed functions. Some few functions seem (!) to have no need for export, so they are excluded from exporting right now.

Maybe we want to export all functions. That whole setup is quite a thicket and it is hard to follow which templated functions call which other templated functions and so in the end need exported basic functions.

Then do the tests again with the shared library.

Another possibility would be to
 - Avoid templated functions that call internals (haha)
 - Make the whole ThreadPool a template so that the now exported functions all would be instantiated by the users (wastes spaces)

**[note]**
Most functions are now exported but not all. Hope the tests instantiate all templates in a way that all needed functions are called at least once. On the other hand we could just make all functions exporty.

**[note]**
ThreadPool is **_BROKEN_** in the current GUL release because the symbols are missing. The tests should use what users will use to make sure it will work. As the tests did not work with the current shared libraries that is true for any user code.

Fixes: #83

### DIFF
--- a/include/gul14/ThreadPool.h
+++ b/include/gul14/ThreadPool.h
@@ -44,6 +44,7 @@ class ThreadPool;
 
 namespace detail {
 
+GUL_EXPORT
 std::shared_ptr<ThreadPool> lock_pool_or_throw(std::weak_ptr<ThreadPool> pool);
 
 } // namespace detail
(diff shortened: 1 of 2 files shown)


alt-graph:
Awww... 🤦‍♂️ 
Thanks!

alt-graph (approved):

PR #85, opened 2024-03-07
ThreadPool: Correct docs

Finii:
Nothing important, just we forgot to change the type in the Doxygen comments.


alt-graph (approved):
Thanks! :+1:

Issue #80, opened 2024-02-28
Changing the number of threads on a ThreadPool


alt-graph:
We could enhance the `ThreadPool` API with functions to increase or decrease the number of threads at runtime.

PR #86, opened 2024-03-07
Make ThreadPool resizable (i.e. change number of threads at runtime)

Finii:
This is all in all rather lazy and does not allow to jump up and down in number of really running threads.
But I guess it is ok.

One could in principle make it less lazy if the number of threads increases (and try to schedule all pending Tasks).
Not sure if that is needed. I guess not - well, therefor I omitted it.

Maybe look at the diff commit by commit.

Fixes: #80

### DIFF
--- a/src/ThreadPool.cc
+++ b/src/ThreadPool.cc
@@ -73,8 +73,8 @@ ThreadPool::~ThreadPool()
 
     for (auto& t : threads_)
     {
-        if (t.joinable())
-            t.join();
+        if (t.t.joinable())
+            t.t.join();
     }
 }
 
@@ -175,6 +175,17 @@ bool ThreadPool::is_shutdown_requested() const
     return shutdown_requested_;
 }
 
+void ThreadPool::set_max_threads(std::size_t num_threads)
+{
+    if (num_threads == 0 || num_threads > max_threads)
+    {
+        throw std::invalid_argument(
+            cat("Illegal number of threads for thread pool: ", num_threads));
+    }
+    std::lock_guard<std::mutex> lock(mutex_);
+    max_threads_ = num_threads;
+}
+
 std::shared_ptr<ThreadPool> ThreadPool::make_shared(
     std::size_t num_threads, std::size_t capacity)
 {
@@ -207,6 +218,18 @@ void ThreadPool::perform_work()
 
     while (!shutdown_requested_)
     {
+        if (threads_.size() > max_threads_)
+        {
+            auto const new_end = std::remove_if(threads_.begin(), threads_.end(),
+                [](Thread& t) {
+                    if (not t.running)
+                        t.t.join();
+                    return not t.running; });
+            threads_.erase(new_end, threads_.end());
+            // Selfdestruct if we are not the survivor
+            if (threads_.size() > max_threads_)
+                break;
+        }
         // mutex is locked
         if (pending_tasks_.empty())
         {
@@ -261,6 +284,11 @@ void ThreadPool::perform_work()
             running_task_ids_.erase(it);
             running_task_names_.erase(running_task_names_.begin() + idx);
         }
+    }
+    // Mark ourselves dead before we stop executing
+    for (auto& t : threads_) {
+        if (std::this_thread::get_id() == t.t.get_id())
+            t.running = false;
     }
 }
 
--- a/src/ThreadPool.cc
+++ b/src/ThreadPool.cc
@@ -52,7 +52,7 @@ ThreadPool::ThreadPool(std::size_t num_threads, std::size_t capacity)
     : max_threads_{ num_threads }
     , capacity_{ capacity }
 {
-    if (num_threads == 0 || num_threads > max_threads)
+    if (num_threads == 0 || num_threads > num_threads_max)
     {
         throw std::invalid_argument(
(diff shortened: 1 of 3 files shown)


tests/test_ThreadPool.cc
```diff
@@ -381,8 +384,16 @@ TEST_CASE("ThreadPool: count_threads()", "[ThreadPool]")
 {
     for (std::size_t i = 1; i <= 2; ++i)
     {
+        std::atomic<bool> stop{ false };
+        // Make sure the pool is removed before the atomic variable go out of scope by defining it after the atomic
```
alt-graph, line 388: Typo: go -> goes (Sorry, the "suggestion" thingie is not available. Only Github knows why.)

alt-graph (commented):
I like the implementation of adding (and especially removing) threads. :+1:

I am not a big fan of "lazy" thread creation, however. One of the big advantages of a pool is that the threads are already there and can pick up work immediately without the overhead of creating a thread. At least it would be nice to have a selector (e.g. `gul14::SpawnThreads::immediately/lazy`) in the constructor and in the `set_max_threads()` call. This way we could leave the choice to the user.

src/ThreadPool.cc
```diff
@@ -175,6 +175,17 @@ bool ThreadPool::is_shutdown_requested() const
     return shutdown_requested_;
 }
 
+void ThreadPool::set_max_threads(std::size_t num_threads)
+{
+    if (num_threads == 0 || num_threads > max_threads)
```
soerengrunewald, line 180: I have not paid much attention to this before. But can we prefix the constant `max_threads` with `ThreadPool::max_threads` to distinguish it from `max_threads_` the class member?

src/ThreadPool.cc
```diff
@@ -262,6 +285,11 @@ void ThreadPool::perform_work()
             running_task_names_.erase(running_task_names_.begin() + idx);
         }
     }
+    // Mark ourselves dead before we stop executing
+    for (auto& t : threads_) {
+        if (std::this_thread::get_id() == t.t.get_id())
```
soerengrunewald, line 290: Isn't this equal to [`not joinable()`](https://en.cppreference.com/w/cpp/thread/thread/joinable)?

soerengrunewald (approved):
Aside from some IMO style thinks it looks good to me

src/ThreadPool.cc
```diff
@@ -61,8 +62,6 @@ ThreadPool::ThreadPool(std::size_t num_threads, std::size_t capacity)
         throw std::invalid_argument(cat("Illegal capacity for thread pool: ", capacity));
 
     threads_.reserve(num_threads);
-    while (threads_.size() < num_threads)
-        threads_.emplace_back([this]() { perform_work(); });
```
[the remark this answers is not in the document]
Finii: Yes, right, because the tasks are started lazily, only on demand
But that will change with Lars's idea of having eager thread creation.

src/ThreadPool.cc
Finii: I guess I don't get it.

Do you mean `t.running` - because when it stopped running it becomes joinable?
I can't really remember, just in the dark, that joinable is not always the same as running.
But reading the docs it seems you are right.
\*pondering\*

Yes, I'll add a commit that simplifies this.

Finii: Ok, my previous comment was of course completely wrong.

We need `Thread` that contains the `std::thread` and the `bool running`, because just from the `std::thread` you can not determine if the thread is doing something or not. From cppref:

image

The thread is joinable immediately after it has been started and is still joinable if it stopped working, until someone joined it. So you can not determine if the thread is executing code or finished executing but waiting to join.

But I guess I just did not understand your question.

Maybe you mean this comment from cppref:

image

So you maybe suggest that we use `joinable()` here instead of the explicit id comparison?

Unfortunately the comment on cppref seems wrong. It just does not work this way and when you compare the two flags you see it controadicts it and we must use the explicit id check.

Here my code / test. Note how this contradicts that enigmatic sentence on cppref.

image

----

So, answering your original question, no, it is not equal; and we must use the explicit id comparsion.

Finii:
> I am not a big fan of "lazy" thread creation, however. One of the big advantages of a pool is that the threads are already there and can pick up work immediately without the overhead of creating a thread. At least it would be nice to have a selector (e.g. `gul14::SpawnThreads::immediately/lazy`) in the constructor and in the `set_max_threads()` call. This way we could leave the choice to the user.

Please have a look at my half-lazy-half-eager implementation idea, described in the git commit message.

* The number of threads given in the ctor is created eagerly.
* The new number of threads given by `set_max_threads()` is created (ramped up) lazily.

alt-graph:
I am neutral on this MR. I do not see a huge benefit in the lazy thread creation, only a marginal one. On the other hand, I think that adding it makes the interface easier to misuse accidentally – people will spin up threads that they did not really want yet and incur penalties for starting up threads they assumed to be there already. If we had more explicit control over this, maybe in the constructor, I could be convinced to a WF. But if @Finii and @soerengrunewald think that this is useful as-is, I won't stand in the way either.

alt-graph (commented):
See above.

PR #89, opened 2024-03-19
ThreadPool::TaskHandle: Add default constructor

alt-graph:
Without a default constructor, it is almost impossible to store a `TaskHandle` in some user-defined data structure. This limits its usefulness dramatically. Hence, this PR adds a simple default constructor and a test to check that the default-constructed task handle does what one would expect (namely, nothing useful).

### DIFF
--- a/include/gul14/ThreadPool.h
+++ b/include/gul14/ThreadPool.h
@@ -133,6 +133,15 @@ public:
     {
     public:
         /**
+         * Default-construct an invalid TaskHandle.
+         *
+         * This constructor creates an invalid TaskHandle which has no result and is not
+         * associated with a ThreadPool.
+         */
+        TaskHandle()
+        {}
+
+        /**
          * Construct a TaskHandle.
          *
          * This constructor is not meant to be used directly. Instead, TaskHandles are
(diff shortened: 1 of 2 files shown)


include/gul14/ThreadPool.h
```diff
@@ -228,7 +237,7 @@ class ThreadPool : public std::enable_shared_from_this<ThreadPool>
 
     private:
         std::future<T> future_;
-        TaskId id_;
+        TaskId id_{ 0 };
```
Finii, line 240:
```suggestion
        TaskId id_ = -1; // unit64-max
```

Finii (approved):
Either solution is fine ;)
But I think we should not hand out unrelated Handles for a Thread that will for sure exist.

alt-graph:
> Maybe we want to change this to the number one then, to try to avoid referencing something that is not correct?

There is no possibility of referencing something that is not correct because we have no pointer to a `ThreadPool` at all, hence we cannot ask it about the wrong ID. Without a `ThreadPool`, the integer ID is meaningless.

Finii:
Yes I just noticed that after setting up a test that calls `TaskHandle{}.cancel();`

Merge away ;)

PR #92, opened 2024-07-03
to_number: Allow basic usage of bool

soerengrunewald:
**Why**  
The DOOCS Serverlib uses `to_number` when resorting values in `D_array`. But it breaks if we wanna instantiate an array of booleans.

This commit will add basic support to allow `to_number` been used with `bool`.

**How**  
First make sure the other implementations wont trigger, since `bool` is _arithmetic_, _integral_ and _unsigned_. Then add a very simple implementation which allows only `0|1` and `true|false`. And to round things up, add some very basic tests.

### DIFF
--- a/include/gul14/to_number.h
+++ b/include/gul14/to_number.h
@@ -487,21 +487,21 @@ inline optional<NumberType> strtold_wrapper(gul14::string_view str) noexcept
  * \since GUL version 1.7 the NAN and INF floating point conversion
  */
 // Overload for unsigned integer types.
-template <typename NumberType,
-          std::enable_if_t<std::is_integral<NumberType>::value
-                           and std::is_unsigned<NumberType>::value
-                           and not std::is_same<NumberType, bool>::value, int> = 0>
-constexpr inline optional<NumberType> to_number(gul14::string_view str) noexcept
+template <typename NumberType>
+constexpr inline std::enable_if_t<std::is_integral<NumberType>::value and
+                                  std::is_unsigned<NumberType>::value,
+                                  optional<NumberType>>
+to_number(gul14::string_view str) noexcept
 {
     return detail::to_unsigned_integer<NumberType>(str);
 }
 
 // Overload for signed integer types.
-template <typename NumberType,
-          std::enable_if_t<std::is_integral<NumberType>::value
-                           and std::is_signed<NumberType>::value
-                           and not std::is_same<NumberType, bool>::value, int> = 0>
-constexpr inline optional<NumberType> to_number(gul14::string_view str) noexcept
+template <typename NumberType>
+constexpr inline std::enable_if_t<std::is_integral<NumberType>::value and
+                                  std::is_signed<NumberType>::value,
+                                  optional<NumberType>>
+to_number(gul14::string_view str) noexcept
 {
     if (str.empty())
         return nullopt;
@@ -530,9 +530,10 @@ constexpr inline optional<NumberType> to_number(gul14::string_view str) noexcept
 }
 
 // Overload for floating-point types.
-template <typename NumberType,
-    std::enable_if_t<std::is_floating_point<NumberType>::value, int> = 0>
-constexpr inline optional<NumberType> to_number(gul14::string_view str) noexcept
+template <typename NumberType>
+constexpr inline std::enable_if_t<std::is_floating_point<NumberType>::value,
+                                  optional<NumberType>>
+to_number(gul14::string_view str) noexcept
 {
     if (str.empty())
         return nullopt;
@@ -565,8 +566,8 @@ constexpr inline optional<NumberType> to_number(gul14::string_view str) noexcept
     return detail::to_unsigned_float<NumberType>(str);
 }
 
-template<typename NumberType, std::enable_if_t<std::is_same<NumberType, bool>::value, int> = 0>
-constexpr inline optional<NumberType> to_number(gul14::string_view str) noexcept
+template<>
+constexpr inline optional<bool> to_number<bool>(gul14::string_view str) noexcept
 {
     size_t pos{};
     bool value{};
--- a/include/gul14/to_number.h
+++ b/include/gul14/to_number.h
@@ -460,8 +460,13 @@ inline optional<NumberType> strtold_wrapper(gul14::string_view str) noexcept
  *   notation using a small or capital "e" ("12e5", "4.2e1", ".2e-4", "2.E5").
  * * Infinity expressions: (optional minus sign) INF or INFINITY ignoring case.
  * * Not-a-number expressions: (optional minus sign) NAN or NAN(char_sequence) ignoring case.
-     The char_sequence can only contain digits, Latin letters, and underscores.
-     The result is a quiet NaN floating-point value.
+ *   The char_sequence can only contain digits, Latin letters, and underscores.
+ *   The result is a quiet NaN floating-point value.
+ *
+ * <h5>Boolean type</h5>
+ * Recognizes only
+ * * "true" or "false" (case insensitive)
+ * * "1" or "0" (no additional leading zeros allowed)
  *
  * The behavior with surrounding whitespace is *undefined*, so
  * it should be removed before passing input to this function.
@@ -485,6 +490,7 @@ inline optional<NumberType> strtold_wrapper(gul14::string_view str) noexcept
  *
  * \since GUL version 1.6
  * \since GUL version 1.7 the NAN and INF floating point conversion
+ * \since GUL version 2.12 bool conversion
  */
 // Overload for unsigned integer types.
 template <typename NumberType>
@@ -566,24 +572,17 @@ to_number(gul14::string_view str) noexcept
     return detail::to_unsigned_float<NumberType>(str);
 }
 
+// Overload for bool
 template<>
 constexpr inline optional<bool> to_number<bool>(gul14::string_view str) noexcept
 {
-    size_t pos{};
-    bool value{};
-
-    for (; pos < str.length(); ++pos)
-    {
-        if (not std::isdigit(str[pos]))
-            break;
-
-        if (str[pos] != '0')
-            value = true;
+    if (str.length() == 1) {
+        if (str[0] == '1')
+            return true;
+        if (str[0] == '0')
+            return false;
+        return nullopt;
     }
-
-    if (pos == str.length())
-        return value;
-
     if (equals_nocase(str, "true"))
         return true;
 
--- a/include/gul14/to_number.h
+++ b/include/gul14/to_number.h
@@ -488,8 +488,9 @@ inline optional<NumberType> strtold_wrapper(gul14::string_view str) noexcept
  */
 // Overload for unsigned integer types.
 template <typename NumberType,
-          std::enable_if_t<std::is_integral<NumberType>::value &&
-                           std::is_unsigned<NumberType>::value, int> = 0>
+          std::enable_if_t<std::is_integral<NumberType>::value
+                           and std::is_unsigned<NumberType>::value
+                           and not std::is_same<NumberType, bool>::value, int> = 0>
 constexpr inline optional<NumberType> to_number(gul14::string_view str) noexcept
 {
     return detail::to_unsigned_integer<NumberType>(str);
@@ -497,8 +498,9 @@ constexpr inline optional<NumberType> to_number(gul14::string_view str) noexcept
 
 // Overload for signed integer types.
 template <typename NumberType,
-          std::enable_if_t<std::is_integral<NumberType>::value &&
-                           std::is_signed<NumberType>::value, int> = 0>
+          std::enable_if_t<std::is_integral<NumberType>::value
+                           and std::is_signed<NumberType>::value
+                           and not std::is_same<NumberType, bool>::value, int> = 0>
 constexpr inline optional<NumberType> to_number(gul14::string_view str) noexcept
 {
     if (str.empty())
@@ -561,6 +563,20 @@ constexpr inline optional<NumberType> to_number(gul14::string_view str) noexcept
     }
 
     return detail::to_unsigned_float<NumberType>(str);
+}
+
+template<typename NumberType, std::enable_if_t<std::is_same<NumberType, bool>::value, int> = 0>
+constexpr inline optional<NumberType> to_number(gul14::string_view str) noexcept
+{
+    constexpr std::array<gul14::string_view const, 2> yes = { "1", "true" };
+    if (std::find(yes.cbegin(), yes.cend(), gul14::lowercase_ascii(str)) != yes.cend())
+            return { true };
+
+    constexpr std::array<gul14::string_view const, 2> no = { "0", "false" };
+    if (std::find(no.cbegin(), no.cend(), gul14::lowercase_ascii(str)) != no.cend())
+            return { false };
+
+    return {};
 }
 
 /// @}
--- a/include/gul14/to_number.h
+++ b/include/gul14/to_number.h
@@ -568,15 +568,28 @@ constexpr inline optional<NumberType> to_number(gul14::string_view str) noexcept
 template<typename NumberType, std::enable_if_t<std::is_same<NumberType, bool>::value, int> = 0>
 constexpr inline optional<NumberType> to_number(gul14::string_view str) noexcept
 {
-    constexpr std::array<gul14::string_view const, 2> yes = { "1", "true" };
-    if (std::find(yes.cbegin(), yes.cend(), gul14::lowercase_ascii(str)) != yes.cend())
-            return { true };
+    size_t pos{};
+    bool value{};
 
-    constexpr std::array<gul14::string_view const, 2> no = { "0", "false" };
-    if (std::find(no.cbegin(), no.cend(), gul14::lowercase_ascii(str)) != no.cend())
-            return { false };
+    for (; pos < str.length(); ++pos)
+    {
+        if (not std::isdigit(str[pos]))
+            break;
 
-    return {};
+        if (str[pos] != '0')
+            value = true;
+    }
+
+    if (pos == str.length())
+        return value;
+
+    if (equals_nocase(str, "true"))
+        return true;
+
+    if (equals_nocase(str, "false"))
+        return false;
+
+    return nullopt;
 }
 
 /// @}
--- a/tests/test_to_number.cc
+++ b/tests/test_to_number.cc
@@ -88,9 +88,9 @@ TEMPLATE_TEST_CASE("test details: error_in_ulp()", "[to_number]",
 TEST_CASE("to_number(): Integer types", "[to_number]")
 {
     REQUIRE(to_number<bool>("0").value() == false);
-    REQUIRE(to_number<bool>("000").value() == false);
+    REQUIRE(to_number<bool>("000").has_value() == false);
     REQUIRE(to_number<bool>("1").value() == true);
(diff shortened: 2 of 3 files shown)


Finii:
I believe we did model this after, or to mimic, some newer C++ library function as kind of backport.

This seems to be not documented, but I guess it was `std::from_chars()`. At least it shares a lot common features, like being able to work with non-terminated strings, explicit error messaging, and so on.

The Doxygen comment mentions this
```
 * Convert an ASCII string_view into a number.
```
```
 * This means to_number() accepts a subset of C++17's from_chars() input format; where it
 * supports the input format it is modeled close to from_chars().
```

The `from_chars()` (and also `to_chars()`) do not support `bool`; and using `from_chars()` to fill a `bool` variable generated the same unreadable error output as does `to_number<bool>()` ;-D

----

The usecase, a `D_array<bool>`, is that even possible?
And then, it seems it uses `get_value_from_string<>()` which does a) not support `bool` and b) does not use `to_number()`?

I'm confused. Is there a PR for the server library?

soerengrunewald:
> I believe we did model this after, or to mimic, some newer C++ library function as kind of backport.
> 
> This seems to be not documented, but I guess it was `std::from_chars()`. At least it shares a lot common features, like being able to work with non-terminated strings, explicit error messaging, and so on.
> 
> The Doxygen comment mentions this
> 
> ```
>  * Convert an ASCII string_view into a number.
> ```
> 
> ```
>  * This means to_number() accepts a subset of C++17's from_chars() input format; where it
>  * supports the input format it is modeled close to from_chars().
> ```
> 
> The `from_chars()` (and also `to_chars()`) do not support `bool`; and using `from_chars()` to fill a `bool` variable generated the same unreadable error output as does `to_number<bool>()` ;-D

Why should the following not be allowed?:
```c++
#include <charconv>
#include <limits>
#include <string_view>
#include <type_traits>

int main()
{
    std::string_view const str = "1";
    bool res{};

    static_assert(std::numeric_limits<bool>::is_integer);
    static_assert(std::is_integral_v<bool> == true);

    // std::from_chars_result
    //     from_chars( const char* first, const char* last,
    //                 /* integer-type */& value, int base = 10 );
    auto r = std::from_chars(str.data(), str.data() + str.size(), res, 2 /*base*/);

    return 0;
}
```
> 
> The usecase, a `D_array<bool>`, is that even possible? And then, it seems it uses `get_value_from_string<>()` which does a) not support `bool`

Therefor we have clientlib PR 200

> and b) does not use `to_number()`?
> 
> I'm confused. Is there a PR for the server library?

See serverlib PR 223

include/gul14/to_number.h
```diff
@@ -563,6 +565,20 @@ constexpr inline optional<NumberType> to_number(gul14::string_view str) noexcept
...
+    constexpr std::array<gul14::string_view const, 2> yes = { "1", "true" };
+    if (std::find(yes.cbegin(), yes.cend(), gul14::lowercase_ascii(str)) != yes.cend())
+            return { true };
+
+    constexpr std::array<gul14::string_view const, 2> no = { "0", "false" };
+    if (std::find(no.cbegin(), no.cend(), gul14::lowercase_ascii(str)) != no.cend())
+            return { false };
+
+    return {};
+}
```
alt-graph, line 580: In a way, this is very elegant, but it seems to me that we are leaving performance on the table.
- If we throw a long string at the function, it is converted to lowercase twice before we realize that we cannot do anything with it.
- I would assume that "1" and "0" are the default representations that we should optimize for (all other `to_number` overloads deal primarily with digits as well).

So maybe a more low-level implementation would be in order, along the lines of:
```cpp
if (str == "1")
    return true;
else if (str == "0")
    return false;
else if (equals_nocase(str, "true"))
    return true;
else if (equals_nocase(str, "false"))
    return false;
else
    return {};
```
Or even, probably with the minimum amount of work:
```cpp
switch (str.length())
{
case 1:
    if (str.front() == '0')
        return false;
    else if (str.front() == '1')
        return true;
    break;
case 4:
    if (gul14::lowercase_ascii(str) == "true")
        return true;
    break;
case 5:
    if (gul14::lowercase_ascii(str) == "false")
        return false;
    break;
};
return {};
```

tests/test_to_number.cc
```diff
@@ -87,6 +87,12 @@ TEMPLATE_TEST_CASE("test details: error_in_ulp()", "[to_number]",
 
 TEST_CASE("to_number(): Integer types", "[to_number]")
 {
+    REQUIRE(to_number<bool>("0").value() == false);
+    REQUIRE(to_number<bool>("1").value() == true);
+    REQUIRE(to_number<bool>("TRUE").value() == true);
+    REQUIRE(to_number<bool>("false").value() == false);
```
alt-graph, line 93: We could maybe add a few more variations of upper- and lowercase here to make sure that the function does what it is supposed to.

alt-graph (commented):
Thanks for the PR! I think adding support for `bool` makes sense, and it also makes sense to support "true" and "false" in addition to "1" and "0". Our `to_number` is quite distinct to C++17's `from_chars()`, so we should not be bound to its limitations. Since this is GUL, we should probably optimize the implementation a bit, though.

alt-graph:
@Finii, @soerengrunewald: This PR has been idle since July, but I think it is worth pursuing. Any more comments/commits?

include/gul14/to_number.h
Finii: I would call that premature optimization.
(But I must confess I like the `length()` approach quite a lot!)

Finii:
I reopened this PR in the repo and just looked at `to_number.h`, and immediately this looks not ready:
The documentation does not mention `bool` at all.

image

Further a still believe, albeit you both think this is a 'nice addon', it is even a misnomer, because `to_number<bool>` does **NOT** convert anything to a number (this is not C where the 'bool' is an integer with value 0 or not-0).

The only driving force here is convenience for some template work in the server library. With almost the same reasoning you could also incorporate `to_number<string>()` or `to_number<complex<double>>()` ("I want to conveniently read some string into a value").

We should not dilute `to_number()` with featureism.

With other words, I am SA this addition.

tests/test_to_number.cc
Finii: Also tests for the likes of `"TRUELIES"` and `"000"` etc are missing if I'm not mistaken.

And here you can argue, if `"000"` not also `false`?
`to_number<int>("00000") == 0`.

This (the limitation with `bool` to `"0"` and `"1"`) is somehow not consistent.

Finii (changes requested):
Documentation issues at least.

alt-graph:
> Further a still believe, albeit you both think this is a 'nice addon', it is even a misnomer, because `to_number<bool>` does **NOT** convert anything to a number (this is not C where the 'bool' is an integer with value 0 or not-0).
> 
> The only driving force here is convenience for some template work in the server library. With almost the same reasoning you could also incorporate `to_number<string>()` or `to_number<complex<double>>()` ("I want to conveniently read some string into a value").
> 
> We should not dilute `to_number()` with featureism.
> 
> With other words, I am SA this addition.

I do not think it is a misnomer. In C++, `is_integral<bool>` is `true`, so in some way `bool` is the shortest available integer. Doubles have string-y representations like "nan" and "inf", too. That does not strike me as a bad mismatch with 0/1/true/false.

Also, I agree with @soerengrunewald that this extension would be *useful:* We have just hand-coded the exact same thing twice in taskolib&friends.

That said, I'll give this a WF.

Finii:
Hmm, why is `0` and `1` handled at all (and not `000`), the stdlib does not:

Screenshot 2024-10-01 at 10 16 28

_Code inspired by https://github.com/taskolib/taskolib/blob/main/src/deserialize_sequence.h_

Finii:
Regarding the one and zero...

Example
```c++
auto x = static_cast<bool>(2); // works
auto x = *to_number("2"); // does not
```

The feeling is strange, I would either omit numbers as valid inputs (bools are in fact neither 1 nor 0); or handle it in the usual way where just zero is false and everything else (which is a number?) is true.

soerengrunewald:
So what is missing to go forward?

soerengrunewald:
How about:
```c++
    if (str.length() == 1) // constexpr
    {
        if (str.front() == '0')
            return false;
        if (str.front() == '1')
            return true;
    }
    else
    {
        if (gul14::equals_nocase(str, "true"))  // constexpr
            return true;

        if (gul14::equals_nocase(str, "false")) // constexpr
            return false;

        auto const number = to_number<int>(str); // constexpr
        if (number.has_value())
        {
            if (number.value() == 0)
                return false;
            if (number.value() == 1)
                return true;
        }
    }
```

Finii:
> How about:

That does look nice.

Still the question for me is
* Is `"2"` not also _**True**_?
* If `"2"` is not a valid input, why is `"1"` valid?

I believe (if we introduce this), we should either stick to the real 'bool' representations "True"/"False"; or if we simulate the C integer-to-bool understanding that 0 is false and everything else is true, well, we should in fact do THAT and not limit it to 1.

What is the reason for this "only 1 is true" idea that you both seem to share?

Finii:
> auto const number = to_number<int>(str); // constexpr

The comment "constexpr" is ... why?

Whatever, of course, we could / should in fact roll a special version to check if
* The input string contains only `'0'` characters (False)
* The input string contains only digits `"0123456789"` (True)
* The input is invalid

I guess that is even faster than a full blown `to_number<int>()`.

Finii:
> The DOOCS Serverlib uses to_number when resorting values in D_array

How does the library serialize a `D_array<bool>` ? I guess internally it is not a vector-of-bool but a vector of integers, and so it is integers in the save-file? Maybe that is the reason why you solely concentrate on `"1"`?

soerengrunewald:
> > auto const number = to_number(str); // constexpr
> 
> The comment "constexpr" is ... why?

To indicate that these calls are constexpr capable.

> 
> Whatever, of course, we could / should in fact roll a special version to check if
> 
>     * The input string contains only `'0'` characters (False)
> 
>     * The input string contains only digits `"0123456789"` (True)
> 
>     * The input is invalid
> 

I'm open to any suggestions.

> 
> I guess that is even faster than a full blown `to_number<int>()`.

Is performance really an issue?

Finii:
> > I guess that is even faster than a full blown `to_number<int>()`.
> 
> Is performance really an issue?

I believe no, but

> Well, I think the performance aspect

And I just wanted to point out that the 'real C' sense of boolean number interpretations is even faster than the "correct" one that allows only one as True.

> I'm open to any suggestions.

And still you do not comment on why "2" is invalid and/or if that would be good or bad or whatever ;)

soerengrunewald:
> > How about:
> 
> That does look nice.
> 
> Still the question for me is
> 
>     * Is `"2"` not also _**True**_?
> 
>     * If `"2"` is not a valid input, why is `"1"` valid?
> 
> 
> I believe (if we introduce this), we should either stick to the real 'bool' representations "True"/"False"; or if we simulate the C integer-to-bool understanding that 0 is false and everything else is true, well, we should in fact do THAT and not limit it to 1.
> 
> What is the reason for this "only 1 is true" idea that you both seem to share?

The main reason is simplicity. Anyhow how about this:
```c++
    if (str.length() == 1 and isdigit(str.front()))
    {
        if (str.front() == '0')
            return false;
        return true;
    }
    else
    {
        if (gul14::equals_nocase(str, "true"))  // constexpr
            return true;
        if (gul14::equals_nocase(str, "false")) // constexpr
            return false;

        auto const number = to_number<int>(str); // constexpr
        if (number.has_value())
        {
            if (number.value() == 0)
                return false;
            return true;
        }
    }
```

Finii:
```c++
gul14::optional<bool> to_bool(gul14::string_view input) {
    bool ret{ };
    size_t i{ };
    for (; i < input.length(); ++i) {
        if (input[i] == '0') {
            continue;
        }
        if (isdigit(input[i])) {
            ret = true;
            continue;
        }
        break;
    }
    fprintf(stderr, "i = %d\n", i);
    if (i == input.length())
        return ret;
    if (gul14::equals_nocase(input, "true"))  // constexpr
        return true;
    if (gul14::equals_nocase(input, "false")) // constexpr
        return false;
    fprintf(stderr, "NIX\n");
    return gul14::nullopt;
}

TEST_CASE("to_number(): bool", "[to_number]")
{
    auto a = to_bool("1");
    REQUIRE(*a == true);
    a = to_bool("002");
    REQUIRE(*a == true);
    a = to_bool("000");
    REQUIRE(*a == false);
    a = to_bool("falSe");
    REQUIRE(*a == false);
    a = to_bool("TruE");
    REQUIRE(*a == true);
    a = to_bool("TruhE");
    REQUIRE(a.has_value() == false);
    a = to_bool("00012345h");
    REQUIRE(a.has_value() == false);
}
```
_Example code but no template_

I believe the special case with `length == 1` is not much better than that, and it is faster in other numbers; not that I believe that performance matters much here.

soerengrunewald:
Do we wanna allow `0x0001` or `0b0001` as well?

Finii:
> Do we wanna allow `0x0001` or `0b0001` as well?

```
  REQUIRE( to_number<signed char>("0x8F").value() == 127 )
due to unexpected exception with message:
  bad optional access
```
_`to_number()` can only work with decimal (base 10)_

alt-graph:
> I believe (if we introduce this), we should either stick to the real 'bool' representations "True"/"False"; or if we simulate the C integer-to-bool understanding that 0 is false and everything else is true, well, we should in fact do THAT and not limit it to 1.
> 
> What is the reason for this "only 1 is true" idea that you both seem to share?

IMO it is sufficient to think of this as the counterpart of some reasonable serializer. For a serializer, reasonable output for a boolean would be
- `true` (+case variations)
- `false` (+case variations)
- `0`
- `1`.
Everything else like `001` or `2` strikes me as an oddity. Granted, one could argue for the `001` case because all other unsigned integers allow leading zeros as well. But `2` is strictly outside of the conversion domain (i.e. of the integer range). We also do not allow `to_number<unsigned short>(200'000)`. The same goes for `-1`, which is also outside of the range.

include/gul14/to_number.h
```diff
@@ -488,17 +488,19 @@ inline optional<NumberType> strtold_wrapper(gul14::string_view str) noexcept
...
     return detail::to_unsigned_integer<NumberType>(str);
 }
 
 // Overload for signed integer types.
 template <typename NumberType,
-          std::enable_if_t<std::is_integral<NumberType>::value &&
-                           std::is_signed<NumberType>::value, int> = 0>
+          std::enable_if_t<std::is_integral<NumberType>::value
+                           and std::is_signed<NumberType>::value
+                           and not std::is_same<NumberType, bool>::value, int> = 0>
```
alt-graph, line 503: The test for bool here is unnecessary because `is_signed<bool> == false` (that said, it does not really hurt either).

Finii, line 503: This change is superfluous. `bool` is not-is-signed.

include/gul14/to_number.h
```diff
@@ -563,6 +565,33 @@ constexpr inline optional<NumberType> to_number(gul14::string_view str) noexcept
...
+{
+    size_t pos{};
+    bool value{};
+
+    for (; pos < str.length(); ++pos)
+    {
+        if (not std::isdigit(str[pos]))
+            break;
+
+        if (str[pos] != '0')
```
alt-graph, line 579: This treats `2` as true, which is inconsistent with the behavior for other integral types. For instance, `to_number<std::uint8_t>(256)` is `nullopt` because the number exceeds the range of the integer type.

include/gul14/to_number.h
```diff
@@ -563,6 +565,33 @@ constexpr inline optional<NumberType> to_number(gul14::string_view str) noexcept
     return detail::to_unsigned_float<NumberType>(str);
 }
 
+template<typename NumberType, std::enable_if_t<std::is_same<NumberType, bool>::value, int> = 0>
+constexpr inline optional<NumberType> to_number(gul14::string_view str) noexcept
```
Finii, line 569: As this code is only intended for one type (i.e. `bool`), should that not just be a template specialization?

Like, 

```diff
--- a/include/gul14/to_number.h
+++ b/include/gul14/to_number.h
@@ -487,21 +487,21 @@ inline optional<NumberType> strtold_wrapper(gul14::string_view str) noexcept
  * \since GUL version 1.7 the NAN and INF floating point conversion
  */
 // Overload for unsigned integer types.
-template <typename NumberType,
-          std::enable_if_t<std::is_integral<NumberType>::value
+template <typename NumberType>
+constexpr inline std::enable_if_t<std::is_integral<NumberType>::value
                            and std::is_unsigned<NumberType>::value
-                           and not std::is_same<NumberType, bool>::value, int> = 0>
-constexpr inline optional<NumberType> to_number(gul14::string_view str) noexcept
+                           , optional<NumberType>>
+                           //and not std::is_same<NumberType, bool>::value, optional<NumberType>>
+to_number(gul14::string_view str) noexcept
 {
     return detail::to_unsigned_integer<NumberType>(str);
 }
 
 // Overload for signed integer types.
-template <typename NumberType,
-          std::enable_if_t<std::is_integral<NumberType>::value
-                           and std::is_signed<NumberType>::value
-                           and not std::is_same<NumberType, bool>::value, int> = 0>
-constexpr inline optional<NumberType> to_number(gul14::string_view str) noexcept
+template <typename NumberType>
+constexpr inline std::enable_if_t<std::is_integral<NumberType>::value
+                           and std::is_signed<NumberType>::value, optional<NumberType>>
+to_number(gul14::string_view str) noexcept
 {
     if (str.empty())
         return nullopt;
@@ -530,9 +530,9 @@ constexpr inline optional<NumberType> to_number(gul14::string_view str) noexcept
 }
 
 // Overload for floating-point types.
-template <typename NumberType,
-    std::enable_if_t<std::is_floating_point<NumberType>::value, int> = 0>
-constexpr inline optional<NumberType> to_number(gul14::string_view str) noexcept
+template <typename NumberType>
+constexpr inline std::enable_if_t<std::is_floating_point<NumberType>::value, optional<NumberType>>
+to_number(gul14::string_view str) noexcept
 {
     if (str.empty())
         return nullopt;
@@ -565,8 +565,9 @@ constexpr inline optional<NumberType> to_number(gul14::string_view str) noexcept
     return detail::to_unsigned_float<NumberType>(str);
 }
 
-template<typename NumberType, std::enable_if_t<std::is_same<NumberType, bool>::value, int> = 0>
-constexpr inline optional<NumberType> to_number(gul14::string_view str) noexcept
+template<>
+constexpr inline optional<bool>
+to_number<bool>(gul14::string_view str) noexcept
 {
     size_t pos{};
     bool value{};
```

(undoing some of your additional enable-if's plus changing this from a two-arguments template to a more clean one argument template).

Finii:
I must admit Lars' last comment (well now not the last comment anymore, but you know probably what I mean) convinced me, and I now would say just `true` `false` `1` and `0` is ok ;-)
Sorry, but I think it has deserved some discussion.

include/gul14/to_number.h
Finii: Already brought up by Lars

soerengrunewald:
Damm I have to leave now...but `str == "0|1"` is not constexpr :'(

Finii:
Fixed constexpr problem.

I believe there is some docu missing still, will add that if needed, lets see

include/gul14/to_number.h
Finii: Fixed by dropping the change.

Finii (approved):

Issue #91, opened 2024-07-03
Compiler warning with SmallVector test


Finii:
Just dumping this here to not forget

![image](https://github.com/gul-cpp/gul14/assets/16012374/1bd256e5-e947-445a-beac-fb541fbb25c0)

alt-graph:
Hmmm... I quickly debugged through the test, and it only calls `delete []` on memory that came from `new []`. So there seems to be no obvious smoking gun here, but maybe this needs a little more investigation.

Finii:
Hmm, very strange, I can (reliably) trigger this warning if `libgul` is built and tested as subproject, but when I change directly into the subproject directory and set the build there up identical, the issue is not surfacing :thinking:   
Well, and it only happens if coverage is to be generated.

```shell
$ cd clientlib
$ meson setup build_gcov3 -Db_coverage=true --wrap=forcefallback --wipe -Dtine=disabled
$ ninja -C build_gcov3 test
```

PR #93, opened 2024-07-04
SmallVector: Make destructor more robust

Finii:
**[why]**
When the library is compiled with heavy instrumentalization (e.g. gcov) in some cases we get this compiler warning:

```
include/gul14/SmallVector.h:426:13: warning: ‘void operator delete [](void*)’ called on unallocated object ‘v3’ [-Wfree-nonheap-object]
  426 |             delete[] data_ptr_;
      |             ^~~~~~~~~~~~~~~~~~
```

The problem is the detection if the thing that currently holds the data is allocated or not. At the moment the size (capacity) is used as indirect marker. It seems that this can go wrong.

**[how]**
Use instead the pointed to address directly. We know the address of the internal data-holding array, and when the current data is not THAT is must be allocated.

Fixes: #91

Reviewed-by: Soeren Grunewald <soeren.grunewald@desy.de>

### DIFF
--- a/include/gul14/SmallVector.h
+++ b/include/gul14/SmallVector.h
@@ -1429,7 +1429,7 @@ private:
     /// Determine if this vector is using allocated external storage.
     bool is_storage_allocated() const noexcept
     {
-        return capacity_ > inner_capacity();
+        return data_ptr_ != internal_array_.data();
     }
 
     /**


alt-graph (approved):
:+1:

PR #94, opened 2024-08-01
join(): Document that string type must provide size()

alt-graph:
We documented that the string type must provide an `operator+=(string&, T)`, but forgot that it must also provide a `size()` member. This PR adds that documentation.


alt-graph:
> 👍
> 
> Both `string` and `string_view` should meet the requirement. And `const char*` is convertible, so how did you notice this requirement?

In [Taskolib](https://github.com/taskolib/taskolib) we have a `Tag` class that is basically a string, but did not provide `size()`. And we wanted to turn a `vector<Tag>` into a string with `gul14::join(tags, " ")`...

PR #95, opened 2024-11-26
SmallVector: Change the SizeType definition to avoid warnings

soerengrunewald:
**Why**
If one does compile with `-Wconversion` enabled, one does get a lot of warnings caused be SmallVector.   
Especially if one does use STL functions like `std::distance` or `std::forward` etc...

**How**
Change the SizeType definition from `uint32_t` to `size_t`.

### DIFF
--- a/include/gul14/SmallVector.h
+++ b/include/gul14/SmallVector.h
@@ -274,7 +274,7 @@ public:
     /// \copydoc ValueType
     using value_type = ValueType;
     /// Unsigned integer type for indexing, number of elements, capacity
-    using SizeType = uint32_t;
+    using SizeType = size_t;
     /// \copydoc SizeType
     using size_type = SizeType;
     /// Signed integer type for the difference of two iterators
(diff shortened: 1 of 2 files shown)


alt-graph:
The 32-bit integer size was chosen deliberately to keep the `SmallVector` object as small as possible. One of the possible use cases we discussed at the time was to have a vector of `SmallVector` objects. A full `size_t` on 64-bit systems also covers way more elements than would make sense for a *small*  vector.

So, in my opinion the way to go would be to add the necessary `static_cast`s to address these warnings, not to change the class.

soerengrunewald:
I discussed with @Finii offline and he also thinks that changing the class is not necessary. Especially since my argument is not necessarily true as well, as proven by a simple test case:
```c++
    SmallVector<double, 3> foo{ 0.1, 11.9, -10.2 };
    REQUIRE(std::distance(foo.begin(), foo.end()) == 3);
```

Finii:
> So, in my opinion the way to go would be to add the necessary `static_cast`s to address these warnings, not to change the class.

100% same :+1:

The problem is when some people always write object construction with parens `()` instead of curlies `{}` where this would not have compiled from the start... Maybe @alt-graph we should have a rule that all object construction in Doocs must use curlies. Makes it more readable too (more obvious if this is a construction or a function call):

```diff
+    using ftype = Field::Vector<T>;
     Field field{ name,
-        Field::Variant(
-            gul14::in_place_type<Field::Vector<T>>, num_elements) };
+        Field::Variant{
+            gul14::in_place_type<ftype>,
+            static_cast<typename ftype::size_type>(num_elements)
+        }
+    };
```

PR #97, opened 2024-11-29
tools: Search for doxygen only if needed

Finii:
When you do not want to generate the documentation Meson still searches for Doxygen - but it will never be used.

**[why]**  
We search for the `doxygen` executable even if we will never use it because the documentation target is not wanted.

**[how]**  
Put the search for the executable one level deeper.


alt-graph (approved):
:+1:

soerengrunewald (approved):
The commit message tag is `tools:` although the change is to `meson`. Aside from that :+1:

Finii:
Change commit tag, no other change. Force push.

PR #98, opened 2024-12-09
Allow mutable function objects in thread pool

alt-graph:
Because of an oversight, one of the `ThreadPool::add_task()` overloads did not accept mutable function objects. This happened because internally a non-mutable wrapper lambda was created that was then in turn unable to call the original mutable function object. This PR modifies the unit tests to check for this and fixes the issue.

### DIFF
--- a/include/gul14/ThreadPool.h
+++ b/include/gul14/ThreadPool.h
@@ -346,7 +346,7 @@ public:
     add_task(Function fct, TimePoint start_time = {}, std::string name = {})
     {
         return add_task(
-            [f = std::move(fct)](ThreadPool&) { return f(); },
+            [f = std::move(fct)](ThreadPool&) mutable { return f(); },
             start_time, std::move(name));
     }
 
--- a/include/gul14/ThreadPool.h
+++ b/include/gul14/ThreadPool.h
@@ -296,6 +296,9 @@ public:
      * // A task with a name
      * pool->add_task([]() { std::cout << "Task 4\n"; }, "Task 4");
      * \endcode
+     *
+     * \since GUL version 2.12.1, add_task() unconditionally accepts mutable function
+     *        objects
      */
     template <typename Function>
     TaskHandle<invoke_result_t<Function, ThreadPool&>>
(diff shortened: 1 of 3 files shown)


include/gul14/ThreadPool.h
```diff
@@ -296,6 +296,9 @@ class ThreadPool : public std::enable_shared_from_this<ThreadPool>
      * // A task with a name
      * pool->add_task([]() { std::cout << "Task 4\n"; }, "Task 4");
      * \endcode
+     *
+     * \since GUL version 2.12.1, add_task() unconditionally accepts mutable function
```
Finii, line 300: Why 'unconditionally', did it accept mutable function objects conditionally before?  
Just, because this can also be a typo because of re-formulation mid sentence.  
(_unconditionally accepts function objects_ vs _accepts mutable FO_)

Finii (approved):
Ah, nice, or rather ugly find!

Me, as functional programming person finds mutable closure strange anyhow ;)

The error message is also not very ... expressive

```
../include/gul14/ThreadPool.h:349:57: error: no match for call to ‘(const ____C_A_T_C_H____T_E_S_T____15()::<lambda()>) ()’
  349 |             [f = std::move(fct)](ThreadPool&) { return f(); },
      |                                                        ~^~
../tests/test_ThreadPool.cc:211:9: note: candidate: ‘____C_A_T_C_H____T_E_S_T____15()::<lambda()> mutable’ <near match>
  211 |         [&last_job]() mutable { last_job = 1; }, now + 120s);
      |         ^
../tests/test_ThreadPool.cc:211:9: note:   passing ‘const ____C_A_T_C_H____T_E_S_T____15()::<lambda()>*’ as ‘this’ argument discards qualifiers
```

I have the feeling we can have similar issues elswhere. `finally`? Will do a quick incomplete search.

_Edit:_

image

include/gul14/ThreadPool.h
alt-graph: Well, there are overloads for `Function()` and `Function(ThreadPool&)`, and only the former was broken. I did not find a better concise formulation for that. :shrug:

PR #99, opened 2025-02-17
Bugfix/resolve conversion warning

soerengrunewald:
With adding more warnings to the DOOCS server library, I notice some warnings, which are not seen when compiling libgul14. It turns out, that we don't compile the tests with warnings enabled.

So this MR, will enable the warnings for the tests as well. And then fix them one by one.

### DIFF
--- a/include/gul14/to_number.h
+++ b/include/gul14/to_number.h
@@ -370,7 +370,7 @@ constexpr inline optional<NumberType> to_unsigned_float(gul14::string_view str) 
     if (not norm_val.has_value())
         return nullopt;
 
-    return detail::pow10(exponent) * *norm_val;
+    return static_cast<NumberType>(detail::pow10(exponent) * *norm_val);
 }
 
 /**
--- a/tests/meson.build
+++ b/tests/meson.build
@@ -50,7 +50,7 @@ tests = [
 test('all',
     executable('libgul-test',
         tests + [ catch_main ],
-        cpp_args : test_cpp_args,
+        cpp_args : [ test_cpp_args, add_cpp_args ],
         dependencies : [ libgul_dep ],
     ),
     timeout : test_time
--- a/tests/test_ThreadPool.cc
+++ b/tests/test_ThreadPool.cc
@@ -614,14 +614,14 @@ TEST_CASE("ThreadPool: Tasks scheduling their own continuation", "[ThreadPool]")
     std::string str;
 
     pool->add_task(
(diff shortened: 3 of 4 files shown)


alt-graph (approved):
:+1:

alt-graph:
@soerengrunewald: Can we get the same patch for GUL17? We should really get moving on that to avoid duplicating everything...

Finii:
> We should really get moving on that

What was our plan?

And I still believe it is bad to switch to `gul17::` in Doocs instead of some more generic `gul::`, where we could upgrade doocs more easily. And it is not that that would raise any problems if doocs depends on `doocs-gul` and gul internally switches from 17 to 20, which (package and GUL) version is used for the doocs libraries is hardwired anyhow.

Or was the problem to have gul14 and 17 in parallel in one server (including doocs libraries)? But then: Why allow that at all. With the doocs way to have version numbers in the library name, there is no need to have 'gul14' or 'gul17' in it at all, because the version < x is gul14 and x < version < y is 17 etc ...

alt-graph:
> > We should really get moving on that
> 
> What was our plan?

My understanding was that you wanted to finish PR https://github.com/gul-cpp/gul17/pull/9 so we can build gul17 Debian packages.

> And I still believe it is bad to switch to `gul17::` in Doocs instead of some more generic `gul::`, where we could upgrade doocs more easily. And it is not that that would raise any problems if doocs depends on `doocs-gul` and gul internally switches from 17 to 20, which (package and GUL) version is used for the doocs libraries is hardwired anyhow.
> 
> Or was the problem to have gul14 and 17 in parallel in one server (including doocs libraries)? But then: Why allow that at all. With the doocs way to have version numbers in the library name, there is no need to have 'gul14' or 'gul17' in it at all, because the version < x is gul14 and x < version < y is 17 etc ...

The thing is that we have to search&replace stuff across our repos anyhow:
a) We do not want to stay with the `gul14` namespace
b) We want to move to `std::string_view`, `std::optional` and so on

So if we search&replace anyhow, we can also make a clean move to a `gul17` namespace. But we can discuss this again if you prefer...

alt-graph:
> @soerengrunewald: Can we get the same patch for GUL17? We should really get moving on that to avoid duplicating everything...

@soerengrunewald, @Finii: Or alternatively, to avoid going through the same pull request twice, I could simply patch the changes into the GUL17 main branch. Would that be OK with you?

PR #100, opened 2025-02-19
ThreadPool: Add get_thread_index()

alt-graph:
# [why]
There are use cases in which it is beneficial to know on which of the *N* pool threads a function gets executed. Sascha recently came up with one such case for the new clientlib NameService.
    
# [how]
Store a thread-local thread index when launching the threads. Return this index when called from a worker thread. For other threads, the thread-local index gets initialized to the maximum value of size_t; if this value is found, get_thread_index() throws an exception.

### DIFF
--- a/include/gul14/ThreadPool.h
+++ b/include/gul14/ThreadPool.h
@@ -113,6 +113,9 @@ public:
     /// A unique identifier for a task.
     using TaskId = std::uint64_t;
 
+    /// A unique identifier for a thread in the pool in the range of [0, count_threads()).
+    using ThreadId = std::vector<std::thread>::size_type;
+
     /**
      * A handle for a task that has (or had) been enqueued on a ThreadPool.
      *
@@ -420,8 +423,9 @@ public:
     std::vector<std::string> get_running_task_names() const;
 
     /**
-     * If called from a worker thread, return the zero-based index of the thread in the
-     * pool (0 for the first thread, 1 for the second, and so on).
+     * Return the thread pool ID of the current thread.
+     *
+     * \returns a thread ID in the range [0, count_threads()).
      *
      * \exception std::runtime_error is thrown if this function is called from a thread
      *            that is not part of the pool.
@@ -429,7 +433,7 @@ public:
      * \since GUL version 2.13
      */
     GUL_EXPORT
-    std::size_t get_thread_index() const;
+    ThreadId get_thread_id() const;
 
     /// Determine whether the queue for pending tasks is full (at capacity).
     GUL_EXPORT
@@ -529,9 +533,9 @@ private:
     /**
      * For worker threads, this is the index of the thread in the threads_ vector.
      * For other threads, the value is meaningless and the variable is initialized to
-     * numeric_limits<size_t>::max().
+     * numeric_limits<ThreadId>::max().
      */
-    thread_local static std::size_t thread_index_;
+    thread_local static ThreadId thread_id_;
 
     /**
      * A condition variable used together with mutex_ to wake up a worker thread when a
--- a/include/gul14/ThreadPool.h
+++ b/include/gul14/ThreadPool.h
@@ -4,7 +4,7 @@
  * \date    Created on November 6, 2018
  * \brief   Declaration of the ThreadPool class.
  *
- * \copyright Copyright 2018-2024 Deutsches Elektronen-Synchrotron (DESY), Hamburg
+ * \copyright Copyright 2018-2025 Deutsches Elektronen-Synchrotron (DESY), Hamburg
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU Lesser General Public License as published
@@ -29,6 +29,7 @@
 #include <condition_variable>
 #include <functional>
 #include <future>
+#include <limits>
 #include <memory>
 #include <mutex>
 #include <stdexcept>
@@ -418,6 +419,18 @@ public:
     GUL_EXPORT
     std::vector<std::string> get_running_task_names() const;
 
+    /**
+     * If called from a worker thread, return the zero-based index of the thread in the
+     * pool (0 for the first thread, 1 for the second, and so on).
+     *
+     * \exception std::runtime_error is thrown if this function is called from a thread
+     *            that is not part of the pool.
+     *
+     * \since GUL version 2.13
+     */
+    GUL_EXPORT
+    std::size_t get_thread_index() const;
+
     /// Determine whether the queue for pending tasks is full (at capacity).
     GUL_EXPORT
     bool is_full() const noexcept;
@@ -514,6 +527,13 @@ private:
     std::vector<std::thread> threads_;
 
     /**
+     * For worker threads, this is the index of the thread in the threads_ vector.
+     * For other threads, the value is meaningless and the variable is initialized to
+     * numeric_limits<size_t>::max().
+     */
+    thread_local static std::size_t thread_index_;
+
+    /**
      * A condition variable used together with mutex_ to wake up a worker thread when a
      * new task is added (or when shutdown is requested).
      */
@@ -583,8 +603,10 @@ private:
     /**
      * The main loop run in the thread; picks one task off the queue and executes it, then
      * repeats until asked to quit.
+     *
+     * \param thread_index  Index of the thread in the threads_ vector
      */
-    void perform_work();
+    void perform_work(std::size_t thread_index);
 };
 
 /**
--- a/src/ThreadPool.cc
+++ b/src/ThreadPool.cc
@@ -135,6 +135,24 @@ std::vector<std::string> ThreadPool::get_running_task_names() const
     return running_task_names_;
 }
 
+ThreadPool::InternalTaskState ThreadPool::get_task_state(const TaskId task_id) const
+{
+    std::lock_guard<std::mutex> lock(mutex_);
+
+    const auto itr = std::find(
+        running_task_ids_.begin(), running_task_ids_.end(), task_id);
+    if (itr != running_task_ids_.end())
+        return InternalTaskState::running;
+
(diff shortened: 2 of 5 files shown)


alt-graph:
BTW: Did you notice this is MR 100? 🍾 🎆 🥳 🎉

src/ThreadPool.cc
```diff
@@ -134,6 +135,16 @@ std::vector<std::string> ThreadPool::get_running_task_names() const
...
 }
 
+std::size_t ThreadPool::get_thread_index() const
+{
+    const std::size_t idx = thread_index_;
+
+    if (idx == std::numeric_limits<std::size_t>::max())
+        throw std::runtime_error("This thread is not part of the pool");
+
+    return idx;
```
soerengrunewald, line 163: Why this extra variable? Also I wonder if the index can actually exceed `max_threads` (or `max_capacity` :thinking:) and with that in mind, shouldn't be the check be:
```c++
    if (thread_index_ > max_threads)
        throw std::runtime_error("This thread is not part of the pool");

    return thread_index_;
```
Also I think this method could easily live in the header and therefor been inline-able by the compiler.

soerengrunewald (approved):
Aside from the minor comment, it looks ok to me, so I will give it a :+1:

src/ThreadPool.cc
alt-graph: Good point about the extra variable, I removed that. As for putting the function into the header, I tried that, but somehow g++ does not like it:
```
/usr/bin/ld: tests/libgul-test.p/test_ThreadPool.cc.o: in function `TLS wrapper function for gul14::ThreadPool::thread_index_':
/home/lfroehli/doocs/library/common/gul14/builddir/../include/gul14/ThreadPool.h:539:(.text._ZTWN5gul1410ThreadPool13thread_index_E[_ZTWN5gul1410ThreadPool13thread_index_E]+0x25): undefined reference to `gul14::ThreadPool::thread_index_'
collect2: error: ld returned 1 exit status
ninja: build stopped: subcommand failed.
```
I guess this is – maybe – a bug in how thread-local storage is implemented in gcc? In any case, inlining is not critical here, so I think we can ignore that.

alt-graph: > Also I wonder if the index can actually exceed max_threads

No, it can't because it is assigned exactly once at the start of the thread, and in the current design we cannot shut down threads at runtime (except by terminating the entire ThreadPool).

Finii:
Hmm, what can I, as a user, do with the `get_thread_index()` value?

The index exposes internals that we probably do not want to expose (apart from other things).

Maybe there is another way to solve Sascha's problem; would be good to know why he thinks he needs the index of the thread in the vector. Which for example assumes there is a order of the threads in the pool, or that the pool is held in some array kind structure etc pp.

src/ThreadPool.cc
Finii: Well of course we wouldn't introduce something that makes it impossible for us to change that in the future.

That would be worse than documenting function return values.

Finii (commented):
This is a new concept, there is no order and no indexes before this change.
It is rather an implementation detail.

Maybe we want to return a `get_thread_Id()` instead? Analogous to `TaskId`. That would leak nothing (even if technically the same value).

But without knowledge of the problem - which has not been described in the PR description - we can not judge what this is all about.

I.e.: Not approve (yet)

alt-graph:
> Hmm, what can I, as a user, do with the `get_thread_index()` value?
> 
> The index exposes internals that we probably do not want to expose (apart from other things).
> 
> Maybe there is another way to solve Sascha's problem; would be good to know why he thinks he needs the index of the thread in the vector. Which for example assumes there is a order of the threads in the pool, or that the pool is held in some array kind structure etc pp.

See https://mcs-gitlab.desy.de/doocs/doocs-core-libraries/clientlib/-/merge_requests/321 for Sascha's merge request. Yes, I guess one could argue that we expose implementation details with `get_thread_index()`... 🤔

alt-graph:
> And IF this is about identifying the threads (see if it is the same or something), why not use `threads::get_id()`.

This is the solution used now by Sascha, but it would be much less complicated if one could just ask the pool about a thread index.

> I'm am completely in the dark. What would one need the ... (implementation details that I cant access from the outside anyhow) ... as a user?

In a nutshell, it is about using resources that are bound to one specific thread. Thread #0 should access resource #0, thread #1 should access resource #1, and so on. And injecting the data into the thread via TLS or so does not seem to be an option.

I guess one of the questions is: Does adding `get_thread_index()` stop us from enhancing the API in some way in the future?

Finii:
> This is the solution used now by Sascha, but it would be much less complicated if one could just ask the pool about a thread index.

I can not really see `get_id` in the mentioned PR's diff.
And I guess `auto my_id = std::this_thread::get_id()` is not very complicated?

> I guess one of the questions is: Does adding `get_thread_index()` stop us from enhancing the API in some way in the future?

Why not call it `get_thread_Id()`? Still it seems that the number is used as ID and not as index, and so we do not need to have it assume there is an index and not it's just a unsorted pot full of threads.

But if we call it `get_thread_Id()` we can also return THE thread id (thread::get_id) from that function; and then the user could do that herself instead?

Maybe you can point me to the use case in the PR, it is rather big and has numerous discussions ;-)

alt-graph:
> Maybe you can point me to the use case in the PR, it is rather big and has numerous discussions ;-)

To my understanding, `NameService::get_thread_index()` does more or less what is proposed here, just in a more complicated fashion. This index is then used for accessing entries of the `ldap_` and `hangup_task_` vectors. `ldap_`, specifically, holds OpenLDAP connections that are explicitly thread-unsafe, so each of them must be associated with one of our threads.

include/gul14/ThreadPool.h
```diff
@@ -418,6 +419,18 @@ class ThreadPool : public std::enable_shared_from_this<ThreadPool>
...
+     * If called from a worker thread, return the zero-based index of the thread in the
+     * pool (0 for the first thread, 1 for the second, and so on).
+     *
+     * \exception std::runtime_error is thrown if this function is called from a thread
+     *            that is not part of the pool.
+     *
+     * \since GUL version 2.13
+     */
+    GUL_EXPORT
+    std::size_t get_thread_index() const;
```
Finii, line 432: Hmm, Github does not have multiline suggestions, but

```diff
      /**
-     * If called from a worker thread, return the zero-based index of the thread in the
-     * pool (0 for the first thread, 1 for the second, and so on).
+     * Returns the current thread's ID in the thread pool.
+     *
+     * It is a number in the range [0, thread_pool_size). 
      *
       * \exception std::runtime_error is thrown if this function is called from a thread
      *            that is not part of the pool.
      *
      * \since GUL version 2.13
      */
     GUL_EXPORT
-    std::size_t get_thread_index() const;
+    std::size_t get_thread_Id() const;
```

Finii (commented):
Hmm, at least I think we should/could _call_ this ID, even if it is the index.

But then, when I look into `NameService.*` I wonder why this solution has been choosen. A lot pre-created `LdapConnection` objects stored as shared ptr in a ("global") vector. Not even a `std::array` :thinking: Ah this is to not-initialize the ldap connection "often".

Some other idea to solve this:

A `ThreadPool` could also handle resources that the workers usually need (in this case `LdapConnection` objects).
For me that would be the more natural change. Then `NameService` does not need to store the connections itself but offloads that to the thread pool; and the thread in the pool always has its own resource, guaranteed.

The change is a bit tricky because we generally do not know the type of the resource that each thread needs and it is hard(er) to specify. I mean, in `C`-isch land the `ThreadPool` could store unique-ptr to void ;) and the `ThreadPool::ThreadPool()` gets a factory function pointer or similar to produce that resource ptrs.

I can not see the solution clearly, but that would need some experiments, but I have the feeling that such an approach would also make the NameService stuff more clear. It looks a lot like he worked around the problem that thread-pool threads can not have precreated resources, and that seems to be not a very strange request. So the ThreadPool could maybe gain something from the problem and not just bare solve the issue.

:thinking:

include/gul14/ThreadPool.h
alt-graph: We already have a unique task ID with type `ThreadPool::TaskId`, so we would be overloading the term "ID". But if that is less controversial than "index", why not.

Finii: Well, `ThreadPool::TaskId` and `ThreadPool::ThreadId` is clearly different in my view.

---

> that would clearly be another PR

Sure. I guess the 'pressure' to come up with a solution is not so high anymore - you merged the solution based on `this_thread::get_id()` as keys in a mapping if I remember correctly. So we could think what would be better.

I'm fine with either (as long as it is called ~~index~~).

But if we have time we could probably look into the 'associated resources' solution.  
That would also work nicely with a varying number of threads, while the 'get the index/id' is more like a kludge.  
Research into that would take some time-investment; so we should decide if it is worthwhile at all. I could do that (research), but not without some verbal commitment at least. Based on the experience with the make-number-of-threads-dynamic I guess this will take 2 (full) work days.

/shrug

alt-graph: Changed with commit 140f60f8ead25f80fda4c0b0f592ca66c09998ff.

alt-graph: > I could do that (research), but not without some verbal commitment at least. Based on the experience with the make-number-of-threads-dynamic I guess this will take 2 (full) work days.

I guess you should check that with Sascha. As far as I am concerned, this small `get_thread_id()` would be a welcome improvement for both the clientlib and GUL14, and sufficient for now.

alt-graph:
Sorry, I messed up the last commit. It is corrected now, and the checks pass.

Finii (approved):
### ###

alt-graph:
Thanks a lot, @soerengrunewald and @Finii!

PR #101, opened 2025-02-26
Enable new NRVO warning

Finii:
[why]
While RVO is guaranteed and we will get warnings about pessimizations the c++17 new optional NRVO is warned on only on request (except on mcvc).

To get optimal code we of course want to benefit from NRVO optimizations.


soerengrunewald (approved):
I would prefer to split the commit into adding the warning to meson and fixing it.
Aside from that :+1:

Finii:
> I would prefer to split the commit into adding the warning to meson and fixing it.

I'm not sure if anything will be gained with splitting this already small commit into two tiny commits.
Furthermore, if someone later wants to see when Wnrvo has been added and what it meant, that is not in one commit, but well, two, the changes that are needed for optimal return handling ;-)

Whatever ;)

TNX for review

PR #103, opened 2025-06-11
Resolve compiler warnings emited by clang

soerengrunewald:
## Why
There are some really annoying warnings which are spammed if compiling with clang.
Since we use gul includes all over, every include will emit these warnings.


Finii:
My understanding is that GUL14 has been sunsetted and will be removed soon.
I'm not sure any minute spent on GUL14 is justified.
Maybe I'm wrong?

What is the timeline, @alt-graph

PR #104, opened 2026-06-22
CI: Update sync pipeline runner

Finii:
[why]
The sync pipeline fails always.
There is no ubuntu20 runner anymore.

[how]
Be more generic, we just use git commands and do not care for OS details at all.


alt-graph (approved):
:+1:

